PR #11, opened 2020-08-12
Standardize Enterprise (and Cloud) Release Notes

paolaperaza:
Hey all, throwing up a PR here with a slightly new format for release notes now that we're separating these out between Cloud and Enterprise. My goal was to:

- Clearly break down all changes for every patch release, both Enterprise and Cloud (and the date they were released)
- Make "latest" patch release for Enterprise users explicit and easy to find (always at the top)
- Correct AWS install docs to NOT specify an explicit version but to add `<platform-version>` to the helm install command and point them to Enterprise release notes to find the latest patch release themselves (cc: @shabx)
- Updated URLs to new docs links in both Cloud + Enterprise release notes doc
- Abstract Enterprise-only stuff for cloud and visa versa (e.g. remove "Enterprise-only changes" in Cloud)

Questions:
- @vishwas-astro I have "Release Date" for Enterprise patch versions as the day we pushed them to Cloud. For Enterprise-only patches there may only be one date, but if we push 0.18.x to Cloud today how do we define when that patch release is made GA for Enterprise, and where is the source of truth for that date?
- I broke down releases in chronological order, appending patches at the bottom of a section for a minor release. Does that seem good to you guys or should we do `0.16.4` at the top and then `0.16.0` at the bottom?

Lemme know what you all think. FYI we still need to clean up the AKS and GKE install docs somewhat significantly.


vishwas-astro:
> Questions:
> 
> * @vishwas-astro I have "Release Date" for Enterprise patch versions as the day we pushed them to Cloud. For Enterprise-only patches there may only be one date, but if we push 0.18.x to Cloud today how do we define when that patch release is made GA for Enterprise, and where is the source of truth for that date?

Enterprise patch versions may or may not get pushed out to Cloud. For eg: if we now come up with 0.16.5 it is not going to get pushed out to our cloud as we are already at ahead (0.18.0) on cloud. At the same time a month ago, when we cut 0.16.1 and 0.16.2, they were pushed out to cloud as well as it was before 0.17.0. was rolled out.

The date, I would say, is when the images are made available in our helm.astronomer.io repo.


> * I broke down releases in chronological order, appending patches at the bottom of a section for a minor release. Does that seem good to you guys or should we do `0.16.4` at the top and then `0.16.0` at the bottom?

I prefer the latest patch release and its release notes at the top and then 0.16.0 at the bottom.

paolaperaza:
@vishwas-astro Cool. I changed up the order such that release notes for newer patch releases are appended at the top - agreed on that.

Re: dates - yea, the split from Cloud minor releases into Enterprise patch releases was precisely my question. I looked here, then for dates (since our helm.astronomer.io repo was deprecated I think in Jan): https://github.com/astronomer/astronomer/releases

- What is 0.16.2? I can't find any GH issue or milestone for issues there, but I do see a PR to bump Astro UI: https://github.com/astronomer/astronomer/releases/tag/v0.16.2. Right now I don't have it accounted for in these release notes
- It looks like 0.16.3 and 0.16.4 have the same release date here FYI (adjusted in release notes doc accordingly)

If that all looks good feel free to approve and I'll go ahead and merge. Thanks for looking at this.

enterprise/stable/09_resources/01_release-notes.md
```diff
@@ -4,48 +4,78 @@ navTitle: "Release Notes"
...
-### New "Environment Variables" Tab in the Astronomer UI
+Release Date: July 22, 2020
 
-Astronomer v0.16 comes with significant improvements to the experience of setting Environment Variables on Astronomer. Namely, we've introduced a dedicated 'Variables' tab, lessening the density of the "Settings" page and making these configurations accessible to "Trial" users on Astronoomer Cloud.
+- Improvement: Default to eventual consistency in Airflow chart version
+- BugFix: Link to Workspace broken in 'SysAdmin' > 'Users' View
+- BugFix: Prevent ability to create Service Account with the same name in Astro UI and CLI
+- BugFix: Error on `$ astro dev init` and `$ astro version` if not authenticated to Astronomer
+- BugFix: Update intermediate cert on the astronomer.io wildcard to resolve `(60) SSL certificate problem: certificate has expired` in Astro CLI
+- BugFix: Restore ability to dynamically pull namespace with `namespace=default` in [KubernetesPodOperator](https://www.astronomer.io/docs/enterprise/stable/customize-airflow/kubepodoperator/) and KubernetesExecutor Config
```
vishwas-astro, line 26: I thought a lot of the above is going in with 0.16.5 and not 0.16.4. Or am I mistaken?

cloud/stable/06_resources/01_release-notes.md
```diff
@@ -8,9 +8,9 @@ description: "Astronomer Cloud Release Notes."
 
 Release Date: July 27, 2020
```
vishwas-astro, line 10: what about release notes for 0.18.0 cloud release?

cloud/stable/06_resources/01_release-notes.md
```diff
@@ -26,35 +26,55 @@ As reported [here](https://forum.astronomer.io/t/error-http-do-failed-i-o-timeou
...
+- BugFix: Error when deleting a 'Pending' Workspace invite in Astro UI 
+- BugFix: "Deployment Status" bubble in the Astro UI persistently blue/pulsating
+- BugFix: Issue with Extra Capacity resetting every time you change an Env Var
+
+### v0.16.3
+
+Release Date: July 17, 2020
+
+- BugFix: Mismatched rendering when switching between Deployments in the Astro UI
+
```
vishwas-astro, line 77: Can we move 0.16.1 and 0.16.3 above 0.16.0?

paolaperaza: Ah yes - missed this in Cloud release notes. Fixed. Thanks for the catch.

cloud/stable/06_resources/01_release-notes.md
paolaperaza: Got it, sorry about that - thanks again for the catch here. Too many files!

enterprise/stable/09_resources/01_release-notes.md
paolaperaza: Yea, I think we need to double check this list - will remove from this PR and submit another  with 0.16.4/0.16.5 details. Thanks @vishwas-astro !

PR #19, opened 2020-08-16
Enhance IAM Role Integration Doc

paolaperaza:
Opening up a draft PR here to address @sjmiller609's feedback in this GitHub issue: https://github.com/astronomer/issues/issues/1587

The changes I made include:
- A new section that walks users through setting up IAM on Kubernetes prior to touching anything on Astronomer (with specific instructions/external docs links for EKS)
- Specify the need to set `serviceAccountAnnotationKey`

I mostly followed the Zendesk ticket here: https://astronomer.zendesk.com/agent/tickets/2262. It sounds like there are still a few outstanding questions relating to content I did not include:

- Do users need to pass the IAM role as an Airflow Connection to get all of this to work?
- Did we figure out why the role was being applied to some workers but not all?
- Does the role have to specify `"Action": "sts:AssumeRole"`?
- Do we support this feature on Azure? Looking at the [annotation in Houston](https://github.com/astronomer/houston-api/blob/main/config/default.yaml#L576) I only see a line for EKS and GKE.

@vparekh94 @ianstanton @sjmiller609 I'm considering this a draft PR, but would love initial feedback from any of you guys since it sounds like you've all been in close touch with Chegg on this and I may not be understanding 100% of that exchange. I'll continue to keep an eye out on that ZD ticket regardless.

### DIFF
--- a/enterprise/v0.16/05_customize-airflow/06_integrate-iam.md
+++ b/enterprise/v0.16/05_customize-airflow/06_integrate-iam.md
@@ -6,13 +6,13 @@ description: "How to append IAM roles to an Airflow Deployment on Astronomer."
 
 ## Overview
 
-As of Astronomer v0.15, IAM roles can now be appended to all pods within any individual Airflow Deployment on the platform.
+On Astronomer, IAM roles can be appended to all Kubernetes pods within any individual Airflow Deployment on the platform.
 
 IAM roles on [AWS](https://aws.amazon.com/iam/faqs/), [GCP](https://cloud.google.com/iam/docs/overview) and other platforms are often used to manage the level of access a specific user (or object, or group of users) have to some resource (or set of resources). The resource in question could be an S3 bucket or Secret Backend, both of which are commonly used in tandem with Airflow and Astronomer and can now be configured to be accessible only to a subset of Kubernetes pods within your wider Astronomer cluster.
 
 To support this functionality, each pod in an Astronomer namespace (each of which maps to an Airflow deployment) is launched by a [Kubernetes Service Account](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/). By annotating the service account with the role, all pods launched will inherit that role.
 
-At a high-level, you can now:
+At a high-level, you can:
 
 * Create an IAM role defining access to the target service or set of services (e.g. AWS S3)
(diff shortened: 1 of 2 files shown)


vparekh94:
Thanks @paolaperaza - going to wait to review this until we get to a resolution with chegg to make sure we don't miss a step!

ianstanton:
@paolaperaza We confirmed with Chegg that they were able to resolve the issue, but they stated the service account annotation step was not necessary. @vparekh94 and I will need to test this out before reviewing.

paolaperaza:
Awesome, thanks for keeping me posted and working with Chegg through this @vparekh94 @ianstanton! Lemme know what all you all find/gimme feedback on this PR whenever you're able to test it.

enterprise/v0.16/05_customize-airflow/06_integrate-iam.md
```diff
@@ -31,32 +31,115 @@ More specific guidelines below.
 ## Prerequisites
 
 * [The Astronomer CLI](/docs/enterprise/v0.16/develop/cli-quickstart/)
-* Your IAM role `arn`
 * Admin access on an Astronomer Workspace
 * Permission to run `$ kubectl describe po` in your cluster
```
vparekh94, line 35: We should generically make this "direct access to your kubernetes cluster"  - feels a little more general in case additional steps need to be taken

vparekh94 (approved):
@paolaperaza I think this is pretty much good to ship. We might want to change some wording to make it play nicely with GKE (and eventually Azure) down the line, but given we're still seeing issues here with AWS, I'm okay to not advertise those two clouds for now.

enterprise/v0.16/05_customize-airflow/06_integrate-iam.md
paolaperaza: Good call - changed this!

paolaperaza:
Awesome, thanks for reviewing this @ianstanton + @vparekh94. Definitely agree with making this doc more multi-provider friendly - for now a lot of it is specific to AWS but I'll continue to keep an eye on this as we ourselves do more testing and get more customers integrating with GKE.

Merging! As you get more info on the AWS annotation being needed/working, let me know and we'll update this doc.

PR #24, opened 2020-08-26
Update Azure install guide

schnie:
Just a quick update - seems like things have changed.

### DIFF
--- a/enterprise/next/03_install/03_azure/01_install-azure-standard.md
+++ b/enterprise/next/03_install/03_azure/01_install-azure-standard.md
@@ -99,7 +99,7 @@ $ az aks show --resource-group <my-resource-group> --name <my_cluster_name> --qu
 Create a static IP in your infrastructure resource group:
 
 ```
-$ az network public-ip create --resource-group <infrastructure-resource-group-name> --name astro-ip --allocation-method static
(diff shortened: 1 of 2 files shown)


schnie:
@petedejoy whats the protocol for a little fix like this? Do we backport these types of things?

paolaperaza:
Thanks for this @schnie. Does this correction apply to previous versions of the platform? I'd make the change you committed to the same file in the `next` folder - it's not published right now, but that's the one that'll eventually turn into the docs folder for the next LTS release.

Docs about new features etc. don't get backported and I'd say little fixes/typos don't _have_ to be backported to earlier versions but can be for general goodwill.

schnie:
@paolaperaza I think it'd apply to any version past or future, at this point in time. @ianstanton what you think?

ianstanton:
@schnie agree it would apply to past, present and future releases.

paolaperaza:
@ianstanton @schnie Ok! For the Azure install doc specifically I just added this to the file in `next` and I think since it's an install doc and people shouldn't be installing an earlier version of Astro I'll stop there.

(We also recently cleaned up this doc and looks like we only applied those changes to 0.16 + next). I'll merge this! Thanks again.

PR #27, opened 2020-09-03
fix syntax typo

ryanahamilton:
image


paolaperaza (approved):
Good call, thanks @ryanahamilton ! FWIW I do think `$ astro help` without the `--` still works but your corrections are correct per CLI output so let's roll with it 🤷‍♀️

ryanahamilton:
You're right! I wasn't paying enough attention beyond the "Error". Yeah, I suppose it would be better to align w/ the CLI output—that's what drew my attention in the first place.

PR #30, opened 2020-09-04
Enterprise patch upgrade instructions

paolaperaza:
This PR addresses this GH Issue: https://github.com/astronomer/issues/issues/1633

I updated our existing "Upgrade Astronomer" doc ([lives here](https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/upgrade-astronomer/)) with:

- An overview of our upgrade policy
- Explanation of platform versioning
- Instructions for patch upgrades, adopted from: https://github.com/astronomer/astronomer/tree/master/bin/migration-scripts#patch-version-updates

The README in the repo I linked above has a lot more info in terms of checking to make sure pod are up, instructions for minor upgrades etc. I wasn't sure what all to include but decided to start narrow to make sure we're clear/accurate and then add info as needed.

@sjmiller609 Since I believe you wrote up those instructions, can you give this a look and make sure it's accurate? A few specific questions:

- Is there a step we can add for customers to "confirm" the patch upgrade was successful?
- Is the script you provided supposed to be prescriptive or do users need to modify it further somehow?
- Any suggestions on language around Airflow chart changes? Will patch releases by definition not include changes to the airflow chart?
- Is this still accurate given eventual consistency?
> The patch upgrade process will NOT affect running tasks if `upgradeDeployments.enabled=false`

### DIFF
--- a/enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
+++ b/enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
@@ -87,6 +87,7 @@ helm3 repo update
 # restart.
 helm3 upgrade --namespace $NAMESPACE \
             -f ./config.yaml \
+            --reset-values \
             --version $ASTRO_VERSION \
             --set astronomer.houston.upgradeDeployments.enabled=false \
             $RELEASE_NAME \
--- a/enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
+++ b/enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
@@ -36,9 +36,10 @@ A few notes before you get started:
 
 Read below for specific guidelines.
 
-> **Note:** Astronomer v0.16.5 and beyond includes "Eventual Consistency" functionality that allows for Airflow Deployments to remain unaffected through a platform upgrade that includes changes to [the Airflow Chart](https://github.com/astronomer/airflow-chart).
+> **Note:** Astronomer v0.16.5 and beyond includes an improved upgrade process that allows Airflow Deployments to remain unaffected through a platform upgrade that includes changes to [the Airflow Chart](https://github.com/astronomer/airflow-chart).
 >
 > Now, Airflow Chart changes only take effect when another restart event is triggered by a user (e.g. a code push, Environment Variable change, resource or executor adjustment, etc).
+
 ### Ensure you have a copy of your Astronomer `config.yaml`
 
 First, ensure you have a copy of the `config.yaml` file of your platform namespace if you don't already.
--- a/enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
+++ b/enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
@@ -6,7 +6,7 @@ description: "How to upgrade the Astronomer Enterprise Platform."
 
 ## Overview
 
-As of Astronomer v0.16, Astronomer releases will be made generally available to Enterprise customers on a quarterly basis within a long-term support (LTS) release model. Critical security and bug fixes will be regularly backported to the latest supported Enterprise LTS version as one or more patch releases. Patch releases will be made available _between_ quarterly LTS releases and require a simple upgrade process.
+As of Astronomer v0.16, Astronomer releases will be made generally available to Enterprise customers on a quarterly basis as part of a long-term support (LTS) release model. Critical security and bug fixes will be regularly shipped as patch versions that follow LTS releases. Patch releases will be made available _between_ quarterly LTS releases and require a simple upgrade process.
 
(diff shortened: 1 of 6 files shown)


enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -4,13 +4,98 @@ navTitle: "Upgrade Astronomer"
...
+A few notes before you get started:
+- The patch upgrade process will NOT affect running tasks if `upgradeDeployments.enabled=false`
+- Patch version updates will NOT cause any downtime to Astronomer services (UI, API, etc.)
+
+Read below for specific guidelines.
+
+> **Note:** Astronomer v0.16.5 and beyond includes "Eventual Consistency" functionality that allows for Airflow Deployments to remain unaffected through a platform upgrade that includes changes to [the Airflow Chart](https://github.com/astronomer/airflow-chart).
+>
+> Now, Airflow Chart changes only take effect when another restart event is triggered by a user (e.g. a code push, Environment Variable change, resource or executor adjustment, etc).
+### Ensure you have a copy of your Astronomer `config.yaml`
```
danielhoherd, line 42: I don't know if ours does, but some renderers treat the lack of a blank line between content and headings differently, so we should probably be consistent with these.
```suggestion

### Ensure you have a copy of your Astronomer `config.yaml`
```

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -4,13 +4,98 @@ navTitle: "Upgrade Astronomer"
...
+
+If you're already on Astronomer's latest minor version (v0.16), you're free to upgrade the platform to a patch release as soon as it's made available.
+
+A few notes before you get started:
+- The patch upgrade process will NOT affect running tasks if `upgradeDeployments.enabled=false`
+- Patch version updates will NOT cause any downtime to Astronomer services (UI, API, etc.)
+
+Read below for specific guidelines.
+
+> **Note:** Astronomer v0.16.5 and beyond includes "Eventual Consistency" functionality that allows for Airflow Deployments to remain unaffected through a platform upgrade that includes changes to [the Airflow Chart](https://github.com/astronomer/airflow-chart).
```
jpweber, line 39: I would suggest we don't put the phrase "eventual consistency" in our docs here. We have used it as jargon internally, but the real meaning of the term is different from what we're describing. I don't want to create confusion with customers that think we mean something else. 

May I suggest something like 
Astronomer v0.16.5 and beyond includes an improved upgrade process that allows Airflow Deployments to remain unaffected through a platform upgrade that includes changes to [the Airflow Chart](https://github.com/astronomer/airflow-chart)

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Good callout, @danielhoherd ! Typo on my part. Fixed.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Good call, @jpweber - wasn't sure on that one. Adjusted!

paolaperaza:
Alright, just applied that feedback - thanks all. @sjmiller609 Let me know if you have any final thoughts on those q's above and approve if this looks good to you.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -4,13 +4,99 @@ navTitle: "Upgrade Astronomer"
...
+
+helm3 repo add astronomer https://helm.astronomer.io
+helm3 repo update
+
+# upgradeDeployments false ensures that Airflow charts are not upgraded when this script is ran
+# If you deployed a config change that is intended to reconfigure something inside Airflow,
+# then you may set this value to "true" instead. When it is "true", then each Airflow chart will
+# restart.
+helm3 upgrade --namespace $NAMESPACE \
+            -f ./config.yaml \
```
sjmiller609, line 89:
```suggestion
            -f ./config.yaml \
            --reset-values \
```

sjmiller609: this is necessary to get the new default values instead of using previous default values.

sjmiller609 (changes requested):

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Done! Thanks @sjmiller609

sjmiller609 (approved):

paolaperaza:
Thanks all. Merging.

PR #33, opened 2020-09-08
AWS Install Guide fix - publicsignups astronomer/issues/#1759

ben-astro:
Hey @paolaperaza just made a change, thanks!


paolaperaza (approved):
Great! Thanks for this @ben-astro - looks good. FWIW I double checked that the syntax here was consistent with our "Manage Platform Users" doc that has instructions on enabling public signups as well - it was 😊 

Link here to that doc for reference: https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/manage-platform-users/

PR #40, opened 2020-09-22
Add details around using non-Lets Encrypt certs in Standard Installation guides - Astronomer/issues/#1833

sunkickr:
Added note to standard install guide as recommended in issue #1833.

### DIFF
--- a/enterprise/next/03_install/01_aws/01_install-aws-standard.md
+++ b/enterprise/next/03_install/01_aws/01_install-aws-standard.md
@@ -118,6 +118,8 @@ sudo kubectl create secret tls astronomer-tls --key /etc/letsencrypt/live/astro.
 
 Make sure to subsitute the appropriate values for your domain.
 
+> **Note:** You can use any SSL Certification you would like by replacing the paths to the Let's Encrypt cert and key .pem files with the paths to your certification's files in the command above.
+
 ## 5. Configure the Database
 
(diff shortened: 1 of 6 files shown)


vparekh94 (changes requested):
@sunkickr this is great - can you include an example of what that command would look like so it's clear for the users? e.g. something like

```bash
kubectl create secret tls astronomer-tls --key <path_to_key> --cert <path_to_cert> --namespace <my-namespace>
```

paolaperaza:
Looks good, @sunkickr - thanks for doing this! My only thought here is that maybe we rephrase the note to be a little bit more actionable and a clear "If, then." So, I might say:

```
 **Note:** If you'd like to use another SSL Certificate authority, replace the paths to the Let's Encrypt cert and key .pem files with the paths to your certification's files in the command above.
```

Tbh I don't know what the "noun" here should be in terms of what to call an alternative to Lets Encrypt, so you guys replace that as needed. In other words, could be, "If you'd like to use another form of SSL Certification..." or "another SSL Certification...".

(and then include an example command like @vparekh94 said).

WDYT?

sunkickr:
@paolaperaza I think the note could be a little more clear too. I like the way you phrased the note. I can make these changes and resubmit the PR.

sunkickr:
Made all changes suggested also found a small typo in the document and updated that too.

paolaperaza (approved):
Awesome, thanks @sunkickr ! Looks good.

vparekh94 (approved):

PR #43, opened 2020-09-24
Manually copy houston-api/config/default.yaml for all documented versions

danielhoherd:
Add reference material from houston-api so we can take that repo private.

I am blocked on the automatic copying of code, but should have that worked out in a few days. This is just to unblock https://github.com/astronomer/issues/issues/824


paolaperaza (approved):
Awesome - looks good to me @danielhoherd !

PR #46, opened 2020-09-28
Remove github houston-api links

danielhoherd:
As part of https://github.com/astronomer/issues/issues/824 and in order to take the houston-api repo private we are removing all references to github houston-api links.

There are two things worth noting though:

1. While I was making changes and verifying that links worked, I found a bunch of anchors that were incorrect in the original docs. This was probably due to the original links not linking to a sha, but to a branch, with let the contents of the file drift over time while the anchors stayed the same. I have corrected some of those instances, but not all of them.

2. There are two links from /cloud that point to /enterprise, because the files did not exist in /cloud yet, and we need a sha to link to. I have included those new files in this PR. The contents of the cloud and enterprise/next files are the same, so the content is good, but we should change the /cloud links to point to cloud in another PR soon, perhaps in a PR to solve the first issue described above. I will open a branch with those changes after this is merged.


paolaperaza (approved):
Alright, ready to rock @danielhoherd ! Clicked through most of the GH links and made sure the versioned files match the versioned folders.

Thanks again for doing this.

PR #48, opened 2020-09-28
Add some js files from houston-api

danielhoherd:
These files are referenced in docs, but I need to merge this change separately from the links so we have a permanent sha to link against, vs a sha that will be squashed or a branch name which will link to content that will change.

### DIFF
--- a/enterprise/next/reference/authorization-handler.js
+++ b/enterprise/next/reference/authorization-handler.js
@@ -1,0 +1,113 @@
+import { hasPermission } from "rbac";
+import log from "logger";
+import { createJWT } from "jwt";
+import prisma from "prisma-client";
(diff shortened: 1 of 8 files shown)


paolaperaza (commented):
Awesome, looks good @danielhoherd - thanks. I noticed `update-deployment-index.js` file is only in v0.16/next folders - that because it's only linked in those? Should we have that file in the rest of the Enterprise folders?

danielhoherd:
@paolaperaza i'm not sure about the update-deployment-index.js file needing to be elsewhere. we can always do editorial improvements later. i'm focused on removing links that are present in the repo so i can move on with other work that requires this to be done.

paolaperaza:
@danielhoherd Ok, fair enough.

paolaperaza (approved):

PR #57, opened 2020-10-09
Add instructions for AC Patch Version upgrades to "Airflow Versioning" doc

paolaperaza:
Hey all - just added some instructions to our existing "Airflow Versioning" doc around upgrading to patch versions of Astronomer Certified. We didn't have _any_ language about AC in here either, so I added some of that framework as well.

@petedejoy @kaxil Would love if you two could take a look at this! Worth noting there are a few small notes that are different between Cloud and Enterprise.

Addresses: https://github.com/astronomer/issues/issues/1912

cc: @vishwas-astro @vparekh94


kaxil (approved):
This is awesome 🚀

petedejoy (approved):
Looks great to me- thanks for the hard work here @paolaperaza!

paolaperaza:
Sweet! Mergin - thanks @petedejoy and @kaxil for reviewing 😊

PR #61, opened 2020-10-17
fix: broken helm links

zacanger:


paolaperaza (approved):
Always great to see our customers in here! @zacanger Thanks so much for pushing up this fix (and catching all the files!) - sorry you ran into that broken link.

Merging 😊

PR #62, opened 2020-10-18
fix: add note about python 3

zacanger:
If Python 3 is installed but not available under the name `python3` I think these would fail, so just adding a note about that.

image


paolaperaza (approved):
Thanks for this one too, @zacanger ! @vparekh94 Can you verify this when you get a change and merge if it looks good to you?

sjmiller609 (approved):

paolaperaza:
Sweet - thanks for reviewing @sjmiller609 and huge thanks again @zacanger! Merging.

PR #66, opened 2020-10-29
Fixing small bug in example DAG on local KubernetesPodOperator doc

kentdanas:
The default .kube/config uses 'docker-desktop' as the cluster context instead of 'docker-for-dekstop'. Updating the example DAG to match so it works without any modifications


paolaperaza (approved):
Nice, thanks @kentdanas !

paolaperaza:
Whoops I merged right as I realized that this change also applies to the corresponding versions of this doc in the Cloud and Enterprise "next" folders.

Committed that change here: https://github.com/astronomer/docs/commit/e28d3c8b36bab761f04319af4327e5d53c504bfa

PR #73, opened 2020-11-18
Update docs to reflect change from DockerHub to Quay

paolaperaza:
Astronomer recently migrated Docker images hosted in DockerHub to Quay in light of Docker's new rate-limit policy:  https://docs.docker.com/docker-hub/download-rate-limit/ 

Given that change, we need to update our docs to:
- [x] Reflect change from DockerHub to Quay as our primary Docker Registry
- [x] Replace all Docker Images to Quay from `astronomerinc` (e.g. `astronomerinc/ap-airflow` > `quay.io/astronomer/ap-airflow`)

### DIFF
--- a/ac/next/01_local.md
+++ b/ac/next/01_local.md
@@ -10,7 +10,7 @@ Designed in close partnership with both Airflow committers and users, Astronomer
 
 There are two primary ways to obtain the Astronomer Certified distribution:
 
-1. [Docker Image](https://hub.docker.com/r/quay.io/astronomer/ap-airflow)
+1. [Docker Image](https://quay.io/astronomer/ap-airflow)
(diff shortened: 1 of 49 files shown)


danielhoherd:
I verified that the changes are good using:
```sh
git diff main.. |
grep '^+[^+]' |
grep -o -E '(https://)?quay.io[^ })`]+' |
sort -u |
grep '^quay' |
xargs -n1 docker manifest inspect >/dev/null
```
and
```sh
git diff main.. | grep '^+[^+]' |
grep -o -E '(https://)?quay.io[^ })`]+' |
sort -u |
grep '^http' |
xargs -n1 curl --max-filesize 1 -s -w "%{http_code}\n" -o /dev/null 
```

danielhoherd (approved):

PR #76, opened 2020-11-18
Various typo fixes

jwitz:
This is a similar PR to the one that closes https://github.com/astronomer/issues/issues/2197, but it covers all docs in all Astronomer versions. I'm not sure what to do with the other one - is there a way for me to delete it?


paolaperaza (approved):
Thanks for these, @jwitz !

PR #80, opened 2020-11-19
Fix HA postgresql example in AKS guide

ianstanton:
Fixed the following in HA postgresql example:
- Indentation
- `synchronousCommit` value needed quotes
- Removed metrics as it required additional values to be set

### DIFF
--- a/enterprise/v0.16/03_install/03_azure/01_install-azure-standard.md
+++ b/enterprise/v0.16/03_install/03_azure/01_install-azure-standard.md
@@ -222,14 +222,12 @@ global:
   postgresqlEnabled: true # Keep True if deploying a database on your AKS cluster.
 
 # Settings for database deployed on AKS cluster.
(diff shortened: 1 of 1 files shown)


paolaperaza (approved):
@ianstanton This looks from a formatting standpoint! Will defer to other folks here to verify technical soundness 🙃 

1 question: Does this have to do with the Postgresql issue on Azure, and if so - for people to leverage this example are they gonna have to be running AC 1.10.14? If so, 1.10.14 will only be avail on the latest 0.16.x patch so it might be worth mentioning that in a note here.

If it has nothing to do with AC/Astronomer versions, looks good.

ianstanton:
@paolaperaza This is our recommended configuration until we have support for Azure Flexible Server. When that support is released (0.23.x?), this section will no longer be necessary. In short - no need for AC 1.10.14 with this configuration.

PR #84, opened 2020-11-24
Edit Cloud VPC access Doc to add new IPs

paolaperaza:
In January 2021, our infrastructure team plans to expand our Astronomer Cloud NAT gateway to support 4 Static IP addresses instead of 1 (https://github.com/astronomer/issues/issues/1871).

We have an existing ["VPC Access" doc](https://www.astronomer.io/docs/cloud/stable/manage-astronomer/vpc-access/) that needs to be updated to reflect this change. This PR:
- Instructs users to allowlist all 4 IP addresses
- Adds a note for existing customers that 3 of the IPs are new but need to be allow-listed by January 2021
- Removes the Ad-Hoc Query section from the doc since that page is deprecated from Airflow as a way to test DB connections

@jpweber Can you take a quick look at this and verify we're instructing users to do the right thing? Most of this doc is an example on AWS Redshift, but I'd love to know if you think we need to expand on anything for clarity. Thank you!

### DIFF
--- a/cloud/stable/05_manage-astronomer/03_vpc-access.md
+++ b/cloud/stable/05_manage-astronomer/03_vpc-access.md
@@ -6,27 +6,34 @@ description: "How to grant Astronomer Cloud access to your VPC."
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


jpweber (approved):
LGTM

PR #89, opened 2020-12-02
Add Config Change Guidelines to Install + Upgrade Docs

jwitz:
Definitely closes https://github.com/astronomer/issues/issues/1883, and potentially closes https://github.com/astronomer/issues/issues/1626. If there's a really popular specific use case that utilizes this kind of configuration (that people might search for in Google :)) let me know!


enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,31 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
 
 We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
 
+## Update Your Astronomer Configuration Without Upgrading
+
+If you want to reconfigure specific variables used on your Astronomer platform, such as resource allocation limits, you can do so without upgrading to a new version.
+
+#### 1. Populate a .yaml file with your current configuration.
+
+Run the following command to create a new file called `config.yaml`:
```
paolaperaza, line 110: Users who are making changes/modifying should already have created a `config.yaml`, but I do agree it's good to give users some guidance if for whatever reason that's not the case.

I'd say something like:

> First, make your config changes to your existing `config.yaml` file as necessary. For an example file that includes all values you could override, refer to our [Houston API source code](https://github.com/astronomer/houston-api/blob/main/config/default.yaml).
> 
> If you're just getting started with Astronomer and don't have that file created yet, refer back to our installation guide for your cloud provider.
> 
> - [AWS](https://www.astronomer.io/docs/enterprise/v0.16/install/aws/install-aws-standard#6-configure-your-helm-chart)
> - [GCP](https://www.astronomer.io/docs/enterprise/v0.16/install/gcp/install-gcp-standard#7-configure-your-helm-chart)
> - [Azure](https://www.astronomer.io/docs/enterprise/v0.16/install/azure/install-azure-standard#6-configure-your-helm-chart)

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,31 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
...
+
+```
+$ helm -n astronomer get values astronomer >config.yaml
+```
+#### 2. Open the new file and reconfigure the variables.
```
paolaperaza, line 115: I'd scrap this section and bring up some of the content in here to the section above. Looks like there are 2 files users could modify:

- https://github.com/astronomer/astronomer/blob/master/values.yaml
- https://github.com/astronomer/houston-api/blob/main/config/default.yaml

Maybe explain the difference between the two and otherwise I think the process here is the same, though I could be missing something. Also - both of these repos will be private soon if not already, so it may be tricky to link to that `/astronomer/astronomer` source code.

We have some files copied over to the docs repo for exactly this reason, so I actually think this is what we should be linking out to - https://github.com/astronomer/docs/tree/main/enterprise/v0.16/reference

enterprise/next/03_install/01_aws/02_install-aws-terraform.md
```diff
@@ -190,27 +190,6 @@ helm ls -n astronomer
...
-    houston:
-      config:
-        email:
-          enabled: true
-          smtpUrl: YOUR_URI_HERE
-
-```
-*Note: You will need to download the helm client to directly call helm commands*
\ No newline at end of file
+Astronomer is deployed on Kubernetes with the [Helm](https://helm.sh) package manager. After completing your installation with Terraform, we recommend updating your configuration through Helm as described in the [Update Your Astronomer Configuration Without Upgrading](https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/upgrade-astronomer/update-your-astronomer-configuration-without-upgrading) topic.
```
paolaperaza, line 195: I'd actually try to be a bit more specific here on when users should use Terraform and when they should make config changes with Helm, since it sounds like it causes a decent amount of confusion.

Expand this section with another sentence or two mayb?

paolaperaza (changes requested):
@jwitz Lookin good here! Made some comments - will look back in here later today so we can wrap this up. Once we're ready, I'd also love @vparekh94 or @ianstanton to look over these changes if either of them have a spare min.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,31 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
 
 We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
 
+## Update Your Astronomer Configuration Without Upgrading
+
+If you want to reconfigure specific variables used on your Astronomer platform, such as resource allocation limits, you can do so without upgrading to a new version.
+
+#### 1. Populate a .yaml file with your current configuration.
```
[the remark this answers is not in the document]
paolaperaza: (If we end up talking about two files here instead of just `config.yaml`, mental note to modify this title then

cloud/stable/04_customize-airflow/04_airflow-alerts.md
```diff
@@ -1,4 +1,4 @@
----
+Deployment-level---
```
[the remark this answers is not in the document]
jwitz: No idea! 0_o Thanks for catching it.

cloud/stable/04_customize-airflow/04_airflow-alerts.md
```diff
@@ -156,6 +156,8 @@ Unlike task-level alerts, deployment-level alerts are sent by Astronomer and do
 | `AirflowMemoryQuota` | Deployment is near its memory quota: it's been using over 95% of it's memory quota for over 10 minutes. |
 
 
+
```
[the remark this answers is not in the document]
jwitz: Definitely - what happened here was I was accidentally working in this branch on some Airflow alerts stuff. To fix that, I copied from `main` back into this branch, so none of this is new content (though it probably appears as such here).

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -47,12 +47,12 @@ First, ensure you have a copy of the `config.yaml` file of your platform namespa
...
 
 ```sh
-$ helm3 get values -n <namespace> <release name of astronomer> > config.yaml
+$ helm3 get values -n <namespace> <release name of astronomer> >config.yaml
 ```
 
 Review this configuration and delete the line `"USER-SUPPLIED VALUES:"` if you see it.
 
-### Verify your current Platform Version
+### Verify Your Current Platform Version
```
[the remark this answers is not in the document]
jwitz: Sounds good - This feature is only available in v0.16+ right? Because that's the only place this topic appears at the moment.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,31 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
 
 We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
 
+## Update Your Astronomer Configuration Without Upgrading
```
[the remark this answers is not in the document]
jwitz: Went even simpler with this: "Update Your Platform Configuration"

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,31 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
...
+
+If you want to reconfigure specific variables used on your Astronomer platform, such as resource allocation limits, you can do so without upgrading to a new version.
```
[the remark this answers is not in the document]
jwitz: Sounds good - Updated the text for this.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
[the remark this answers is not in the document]
jwitz: Ah I see, didn't realize that `config.yaml` use was common across our docs and essential to installation.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Do we still want to leave the helm command in there if they've somehow lost the file, or need it updated with their org's current values? Also, how do we feel about using the word 'organization' for Enterprise customers?

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: I decided to just link out to the last file here and avoided explaining the difference between those repos. And instead of referring to 'source code' (since that's going private), I used the phrase 'default configuration file'.

enterprise/next/03_install/01_aws/02_install-aws-terraform.md
jwitz: I can be more firm with our prescription on using Helm, but based on what Kris told me we really don't want people using Terraform after installation. I'll defer to @vparekh94 or @ianstanton on that though, since I'm definitely not familiar with Terraform usage.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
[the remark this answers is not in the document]
paolaperaza: Good question - I'd think it applies to versions lower than 0.16 but I'm not 100% sure - I'd have @ianstanton double check on this.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,35 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
 
 We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
 
+## Update Your Platform Configuration
+
+You can reconfigure your Astronomer platform without upgrading your Astronomer version using Helm. This is useful for [integrating an Auth system](https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/integrate-auth-system), [adding a registry backend](https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/registry-backend), (changing resource allocation limits)[https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/configure-platform-resources], or any other change that applies to your entire platform.
```
paolaperaza, line 106: I'd add some "umph"/a little more context here here and remove `reconfigure` (sort of implies that you're doing something "again" which isn't really the case) -

> There are a few ways to customize Astronomer to fit the needs of your team. For example, you may choose to:
> - Integrate an Auth System
> - Change resource allocation limits
> - Add a registry backend
>
> To apply the platform configuration changes above and any others that exclude the initial installation and Astronomer version upgrades, follow the steps below.
> 

Sidenote: I think there's an extra space in the `change resource allocation limits` link - can't tell for sure but just double check the url is tied to the description.

Alsoooo I'm a big fan of bullet points/lists to break things up visually if there's every 2+ items on something. WDYT?

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,35 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
...
+
+#### 1. Open your `config.yaml` file.
```
paolaperaza, line 108: Do we want to standardize on adding `.` in H4s? I'd opt to _not_ + I think that's the case in most other docs, but maybe double check and let's make sure we standardize that

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,35 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
 
 We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
 
+## Update Your Platform Configuration
```
paolaperaza, line 104: Still thinking about dis title - I might like `Customize` your Platform Configuration even more actually? Users here are typically coming here to make relatively permanent changes to make x thing work better for them or to leverage x feature or functionality - not so much "updating" stuff.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,35 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
...
+
+This file was created when you installed Astronomer using one of the following guides:
+* [AWS EKS Installation Guide] (https://www.astronomer.io/docs/enterprise/v0.16/install/aws/install-aws-standard#6-configure-your-helm-chart)
+* [GCP GKE Installation Guide] (https://www.astronomer.io/docs/enterprise/v0.16/install/gcp/install-gcp-standard#7-configure-your-helm-chart)
+* [Azure AKS Installation Guide] (https://www.astronomer.io/docs/enterprise/v0.16/install/azure/install-azure-standard#6-configure-your-helm-chart)
+
+If you need to create a new copy of this file, the following Helm command will create a new file called `config.yaml` populated with your organization's configured values::
+```
+$ helm -n astronomer get values astronomer >config.yaml
+```
```
paolaperaza, line 118: Curious to get @ianstanton's thoughts but I actually wouldn't add these instructions here - customers should 100p only be coming to this doc if they've already installed the platform.

If they don't have a `config.yaml` file that means they didn't follow install instructions and should go back there to follow our entire guidelines around getting this file created with the right default values and what not

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,35 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
...
+#### 2. Update key-value pairs in `config.yaml`.
+
+For an example file that includes all of the possible values you can configure, refer to [the default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml).
+
+To update any of your existing key-value pairs, simply modify the existing value in `config.yaml`. To update a setting that hasn't yet been configured, copy the .yaml key-value pair you want to update from the default configuration file into your `config.yaml` file and update the value from there. After copying and modifying the key-value pairs, ensure they have the same order and indentation as they had in the default configuration file.
```
paolaperaza, line 123: > `To update any of your existing key-value pairs, simply modify the existing value in `config.yaml`

This sounds a bit awk. I'd say maybe -

> To override any existing key-value pairs, modify them directly in your `config.yaml` file. To add a new set of values that you don't specify already, copy the corresponding key-value pair from the default configuration file into your own `config.yaml` and modify the value from there.
>
> Once you've changed all necessary key-value pairs, make sure they're in the exact same order and indentation as those listed in the default configuration file we provided above. An indentation error could cause a value to not be properly applied.


GREAT call specifying that users make sure things are indented correctly - you'd be surprised how many people mess that up + subsequently don't see their changes applied once they run helm uprade.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -101,3 +101,35 @@ If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/ent
...
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
```
paolaperaza, line 128: Make sure we standardize spacing - here it looks like there's a full line before the command but in the following instruction the command goes in the line directly under it.

I'd defer to keeping a line in between, so:

> Run this:
>
> `$ blah blah`

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: I think we should definitely use a period for steps, especially if we aren't going to capitalize steps as titles. For other h4 usages though (such as titles) I agree that we shouldn't use periods.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: I think `Customize` is too strong of a word for what users are doing here, which is updating the values of variables which are pre-determined in the system. I associate customization more with personal changes built *on top* of existing infrastructure, for instance if someone could write and deploy their own key-value pairs. Would `Reconfigure Your Platform` work?

jwitz: Or `Configure`, based on your comment above.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: As long as we add an extra bullet point for the default config file, I think bullets work. While I'm usually in agreement about their ability to break things up, I think bulleted lists have a connotation of comprehensively mapping the possible options for something, which is why I didn't originally use it. I think the version in my new commit looks good!

Updated that last sentence to the "Cellar Door" of technical documentation: "To do so:"

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: If they will be more successful by just using the file generated in those installation processes, then I agree we should remove this part.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Done! Agreed on this formatting.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Put my spin on this, but agreed that it needed some tightening up - thanks!

enterprise/next/03_install/01_aws/02_install-aws-terraform.md
paolaperaza: @jwitz Yea, I think for that same reason I think it's worth re-iterating that here just to be _extra_ clear that Enterprise customers should not be using Terraform following the platform install

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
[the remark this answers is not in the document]
paolaperaza: Actually, sorry - yes! This section applies to the "patch upgrade" instructions, which are only available and only apply for 0.16.x - they should also be in the `/next` folder though.

enterprise/next/03_install/01_aws/02_install-aws-terraform.md
jwitz: Just made a commit for this comment in v0.16. Will push to the rest of the versions if you think it's sound!

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -52,12 +53,12 @@ $ helm3 get values -n <namespace> <release name of astronomer> > config.yaml
 
 Review this configuration and delete the line `"USER-SUPPLIED VALUES:"` if you see it.
 
-### Verify your current Platform Version
+### Verify your Current Platform Version
```
paolaperaza, line 56: @jwitz It looks like the instructions to upgrade patch versions in this doc are not numbered and don't have periods. Since we're modifying this doc as a whole, do you want to apply the formatting from the section below here as well to make sure we're standardizing, at least within the same file?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,41 @@ helm3 upgrade --namespace $NAMESPACE \
...
+To update any of your existing settings, modify them directly in `config.yaml`. To update a setting you haven't already specified, copy the corresponding key-value pair from the [default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml) into your `config.yaml` file and modify the value from there.
+
+When you have finished updating the key-value pairs, ensure that they have the same relative order and indentation as they do in the default configuration file. If they don't, your changes might not be properly applied.
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n astronomer upgrade astronomer astronomer/astronomer --version=<your current version> -f config.yaml
```
paolaperaza, line 133: @jwitz I'm looking at our ["Integrate an Auth System" doc](https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/integrate-auth-system) and it looks like the command listed there is:

`$ helm upgrade <platform-release-name> -f config.yaml --version=<platform-version> astronomer/astronomer -n <your-namespace>`

I think the command you have listed does the same thing, but it looks like we should:

1. Specify that `--version=<your current version>` --> `--version=<platform-version>` (users have a lot of versions to deal with, so we should specify "version of what")
2. We may have also assumed `platform-release-name` and `your-namespace`

Maybe verify with @ianstanton real quick on technical accuracy, but if both are accurate maybe just replace the command you have here with the one from the other doc.

paolaperaza: It also looks like other docs have users run:

```
$ helm ls --namespace <your-namespace>
```

Maybe add?

paolaperaza: + consider an example:

> For example, if you're running Astronomer v0.16.10 and your platform release name is `astronomer`, you might run:
> 
> ```
> $ helm upgrade astronomer -f config.yaml --version=0.16.10 astronomer/astronomer -n astronomer
> ```

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,41 @@ helm3 upgrade --namespace $NAMESPACE \
...
+```
+To see the updated key-value pairs in your terminal, you can run:
+
+```
+$ helm3 get values -n <namespace> <release name of astronomer>
```
paolaperaza, line 138:
```suggestion
$ helm3 get values -n <namespace> <platform-release-name>
```

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,41 @@ helm3 upgrade --namespace $NAMESPACE \
...
+* Update any other key-value pair specified in the [default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml)
+
+To do so:
+
+#### 1. Open your config.yaml file.
+
+This file was created when you installed Astronomer using one of the following guides:
+* [AWS EKS Installation Guide](https://www.astronomer.io/docs/enterprise/v0.16/install/aws/install-aws-standard#6-configure-your-helm-chart)
+* [GCP GKE Installation Guide](https://www.astronomer.io/docs/enterprise/v0.16/install/gcp/install-gcp-standard#7-configure-your-helm-chart)
+* [Azure AKS Installation Guide](https://www.astronomer.io/docs/enterprise/v0.16/install/azure/install-azure-standard#6-configure-your-helm-chart)
```
paolaperaza, line 120:
```suggestion
This file was created when you installed Astronomer using one of the following guides:

* [AWS EKS Installation Guide](https://www.astronomer.io/docs/enterprise/v0.16/install/aws/install-aws-standard#6-configure-your-helm-chart)
* [GCP GKE Installation Guide](https://www.astronomer.io/docs/enterprise/v0.16/install/gcp/install-gcp-standard#7-configure-your-helm-chart)
* [Azure AKS Installation Guide](https://www.astronomer.io/docs/enterprise/v0.16/install/azure/install-azure-standard#6-configure-your-helm-chart)
```

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: @jwitz Resolving this convo since I think you covered this change. I do like the term "organization" - don't think it's specific to Enterprise versus Cloud though for the most part.

I also really like using "team" + would prob use that over organization since at the Enterprise level we work with a lot of different _teams_ within "organizations" (companies) - and each team might have different setup/platform installed.

enterprise/next/03_install/01_aws/02_install-aws-terraform.md
paolaperaza: @jwitz Perfect yes just saw your commit - that looks great!

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Great point on the bullet lists - add that to the style guide! Not something I would've been thinking about. I'm cool with your changes here!

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Ah ok, so period for steps but not necessarily all H4s ✅  That sounds good - add to style guide!

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: @jwitz Oh yea that's super fair pushback on the "Customize". How about: `Configure Platform Settings`?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: (Just added a space here)

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Could we do `Configure Platform Settings Between Upgrades`? I'm worried that if we don't mention upgrades in the title, the section will look a little out of place in the `Upgrade Astronomer` section.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Ah yeah, didn't realize these were steps haha. I'll format these in v0.16

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Seems like the commands can be run in any order, since the one I wrote is how Kris ran it on his terminal, but I def think it's best to be consistent across usages. 

I like having the hyphen names for the user-entered variables. How do you feel about always prefixing `your-` to any variable name if it needs to be added by a user?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Did this but added `your-` to the variable name.

enterprise/v0.12/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -30,3 +30,39 @@ Astronomer platform releases follow a semantic versioning scheme. All versions a
...
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n astronomer upgrade astronomer astronomer/astronomer --version=<your current version> -f config.yaml
+```
+To see the updated key-value pairs in your terminal, you can run:
+
+```
+$ helm3 get values -n <namespace> <release name of astronomer>
```
ianstanton, line 67: These commands will vary depending on the minor version they are running. I believe we introduced helm 3 support in 0.15, so if they are on 0.13 or below the commands will slightly differ (I see this doc is on the 0.12 page).

enterprise/v0.13/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -30,3 +30,39 @@ Astronomer platform releases follow a semantic versioning scheme. All versions a
...
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n astronomer upgrade astronomer astronomer/astronomer --version=<your current version> -f config.yaml
+```
+To see the updated key-value pairs in your terminal, you can run:
+
+```
+$ helm3 get values -n <namespace> <release name of astronomer>
```
ianstanton, line 67: These commands will vary depending on the minor version they are running. I believe we introduced helm 3 support in 0.15, so if they are on 0.13 or below the commands will slightly differ (I see this doc is on the 0.13 page).

enterprise/v0.15/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -31,3 +31,38 @@ Astronomer platform releases follow a semantic versioning scheme. All versions a
...
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n astronomer upgrade astronomer astronomer/astronomer --version=<your current version> -f config.yaml
+```
+To see the updated key-value pairs in your terminal, you can run:
+
+```
+$ helm3 get values -n <namespace> <release name of astronomer>
```
ianstanton, line 67: Change `helm3` to `helm`

ianstanton: The command directly above uses namespace and release-name `astronomer`. We can assume the same here and change both `<namespace>` and `<release name of astronomer>` to `astronomer`.

enterprise/v0.15/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -31,3 +31,38 @@ Astronomer platform releases follow a semantic versioning scheme. All versions a
...
+To update any of your existing settings, modify them directly in `config.yaml`. To update a setting you haven't already specified, copy the corresponding key-value pair from the [default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml) into your `config.yaml` file and modify the value from there.
+
+When you have finished updating the key-value pairs, ensure that they have the same relative order and indentation as they had in the default configuration file. If they don't, your changes might not be properly applied.
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n astronomer upgrade astronomer astronomer/astronomer --version=<your current version> -f config.yaml
```
ianstanton, line 62: We should mention here that this example command assumes the `namespace` and `release-name` are both `astronomer`. A customer may have different values.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,46 @@ helm3 upgrade --namespace $NAMESPACE \
...
+To update any of your existing settings, modify them directly in `config.yaml`. To update a setting you haven't already specified, copy the corresponding key-value pair from the [default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml) into your `config.yaml` file and modify the value from there.
+
+When you have finished updating the key-value pairs, ensure that they have the same relative order and indentation as they do in the default configuration file. If they don't, your changes might not be properly applied.
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n <your-namespace> upgrade <your-platform-release name> astronomer/astronomer --version=<your-platform-version> -f config.yaml
```
ianstanton, line 134: We give this example in this `v0.16` doc, but I didn't see it in the others.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,46 @@ helm3 upgrade --namespace $NAMESPACE \
...
+```
+For example, if you're running Astronomer v0.16.10 and your platform release name is astronomer, you might run:
+```
+$ helm upgrade astronomer -f config.yaml --version=0.16.10 astronomer/astronomer -n astronomer
```
ianstanton, line 138: Not the worst thing in the world, but I noticed the `helm` example commands are written in different order. This won't hurt functionality, but it would be good to choose one command format for us to stick with across all docs.

ianstanton (changes requested):

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -52,15 +53,15 @@ $ helm3 get values -n <namespace> <release name of astronomer> > config.yaml
...
 Review this configuration and delete the line `"USER-SUPPLIED VALUES:"` if you see it.
 
-### Verify your current Platform Version
+#### 2. Verify your current platform version.
 
 To verify the version of Astronomer you're currently operating with, run:
 
 ```sh
-helm3 list --all-namespaces | grep astronomer
+$ helm3 list --all-namespaces | grep astronomer
```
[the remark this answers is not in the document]
jwitz: Sounds good, changing this across all instances you mention. I noticed we use "shell session" coloring on this command but not others. Is that appropriate for all helm commands? For reference, here's what it looks like on our website. 
Screen Shot 2020-12-11 at 12 47 55 PM

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -99,5 +101,40 @@ helm3 upgrade --namespace $NAMESPACE \
...
+To update any of your existing settings, modify them directly in `config.yaml`. To update a setting you haven't already specified, copy the corresponding key-value pair from the [default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml) into your `config.yaml` file and modify the value from there.
 
+When you have finished updating the key-value pairs, ensure that they have the same relative order and indentation as they had in the default configuration file. If they don't, your changes might not be properly applied.
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
+```
+$ helm -n astronomer upgrade astronomer astronomer/astronomer --version=<your current version> -f config.yaml
```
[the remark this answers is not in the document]
jwitz: How likely is it that customers will have different values here? If it's anything more frequent than "extremely rare", then it's probably better for us to use pointy brackets with generic values rather than `astronomer` across all uses of the command (ie. `<namespace>` and `<release name of astronomer>`)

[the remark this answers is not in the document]
jwitz: We could simply add a line that says "Note that for most customers, <namespace> and <release name of astronomer> will both be the default value of Astronomer" or something like that

jwitz:
Thanks for the feedback @ianstanton! I should've told you this before, but I'm using v0.16 as the working directory for any changes (just pushed changes there) - after review, I'll push those changes to the rest of the files. Based on your review, I have a few more questions:

* What would the commands look like for v0.12 and v0.13 (that you mentioned would be different)?
* Is it appropriate to use "Shell session" styling on these commands?
* Can we provide more guidance on how customers can find their namespace and platform release name? I took the example out, because I realized it's not easy to parse when both the namespace and release name are "astronomer". It's probably better for us in the long-run to use pointy brackets whenever there's a user-entered value in a command.

enterprise/v0.13/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: What are the specific commands we should use in earlier versions?

enterprise/v0.15/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Resolving based on comments elsewhere.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Resolving based on comments elsewhere

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Nice call, this should be fixed in the latest version!

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -99,5 +101,40 @@ helm3 upgrade --namespace $NAMESPACE \
...
+```
+To see the updated key-value pairs in your terminal, you can run:
+
+```
+$ helm3 get values -n <namespace> <release name of astronomer>
```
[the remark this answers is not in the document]
jwitz: Resolving based on comments elsewhere

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -47,12 +49,12 @@ First, ensure you have a copy of the `config.yaml` file of your platform namespa
 To do this, you can run:
 
 ```sh
-$ helm3 get values -n <namespace> <release name of astronomer> > config.yaml
+$ helm3 get values -n <namespace> <release name of astronomer> >config.yaml
```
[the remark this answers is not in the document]
jwitz: Resolving based on comments elsewhere

enterprise/v0.15/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Decided that this was a confusing example, so updated these commands to only use generic variables. Is there more info we can provide on where they can find their `namespace` and `release-name`, or should those things be self-evident?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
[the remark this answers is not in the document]
ianstanton: @jwitz Yep, you can use shell session coloring with these commands for consistency.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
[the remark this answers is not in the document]
ianstanton: It's not rare for a customer to use different values here.

enterprise/v0.13/06_manage-astronomer/05_upgrade-astronomer.md
ianstanton: The earlier versions (that use helm 2) will not use a `-n` (namespace) flag. Example: `helm get values astronomer -n astronomer` becomes `helm get values astronomer` if using helm 2.

enterprise/v0.15/06_manage-astronomer/05_upgrade-astronomer.md
ianstanton: Namespace _should_ be self-evident. You can list all namespaces on your kubernetes cluster with `kubectl get ns`, but it is up to the customer to know which they used to install Astronomer. Release name can be found with `helm ls -n <namespace>`.

ianstanton:
@jwitz 

> * What would the commands look like for v0.12 and v0.13 (that you mentioned would be different)?

Astronomer v0.12 and v0.13 used Helm 2. In Helm 2, releases were not scoped at the  namespace level. Here are some example command translations (helm 3 -> helm 2):

`helm get values -n <namespace> <release name of astronomer> > config.yaml`
becomes
`helm get values <release name of astronomer> > config.yaml`

`helm list --all-namespaces | grep astronomer`
becomes
`helm list | grep astronomer`

`helm -n <your-namespace> upgrade <your-plarform-release-name> astronomer/astronomer --version=<your-platform-version> -f config.yaml`
becomes
`helm upgrade <your-plarform-release-name> astronomer/astronomer --version=<your-platform-version> -f config.yaml`

> * Is it appropriate to use "Shell session" styling on these commands?

These are shell commands, so I would say it is appropriate. Check the format of similar commands across the website to ensure consistency (If you think we shouldn't be using them _anywhere_, that's fine.)


> * Can we provide more guidance on how customers can find their namespace and platform release name? I took the example out, because I realized it's not easy to parse when both the namespace and release name are "astronomer". It's probably better for us in the long-run to use pointy brackets whenever there's a user-entered value in a command.

This will be determined on initial installation. For example, a user could decide to use `astro-prod-1` as their namespace and `astro` as their release name (`helm install astro -f config.yaml astronomer/astronomer -n astro-prod-1`). A user can list all helm releases on the cluster with `helm ls -A`. They can also list all namespaces on the cluster with `kubectl get ns`. If they forget where they installed Astronomer, they'll be able to look at the pods in each namespace and determine which they used.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -40,27 +41,27 @@ Read below for specific guidelines.
 >
 > Now, Airflow Chart changes only take effect when another restart event is triggered by a user (e.g. a code push, Environment Variable change, resource or executor adjustment, etc).
 
-### Ensure you have a copy of your Astronomer `config.yaml`
+#### 1. Ensure you have a copy of your Astronomer config.yaml file.
```
paolaperaza, line 44: @jwitz How do we want to standardize "steps"? Do we want to list `### Step 1: Ensure...` as we do in Cloud Quickstart? If so, let's make sure that's the case across the board for this doc

paolaperaza: Also, if I'm reading correctly - is this an H4 under an H2?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,45 @@ helm3 upgrade --namespace $NAMESPACE \
 
 If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/enterprise/v0.16/resources/release-notes) from an earlier minor version, submit a request to [Astronomer Support](https://support.astronomer.io).
 
-We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
+We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Winter 2020.
+
+## Configure Platform Settings Between Upgrades
+
+You can configure platform-wide settings without needing to upgrade Astronomer versions. Using Helm, any of the settings your platform uses, including ones available in the Astronomer UI, can be updated through a few simple commands. For example, you can use Helm to:
```
paolaperaza, line 106:
```suggestion
You can configure platform-wide settings without needing to upgrade Astronomer versions. Using Helm, any of the settings your platform uses can be updated through a few simple commands. For example, you can use Helm to:
```

Platform configuration changes aren't really the ones that show up in the Astronomer UI, so I'd say it's weird to call out that example here. I'd just leave it as ^that

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,45 @@ helm3 upgrade --namespace $NAMESPACE \
...
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
+```sh
+$ helm -n <your-namespace> upgrade <your-platform-release-name> astronomer/astronomer --version=<your-platform-version> -f config.yaml
+```
+
+If you don't know the value for `<your-platform-release-name>`, you can find it by running `helm ls -n <your-namespace>`.
```
paolaperaza, line 137:
```suggestion
If you don't know the value for `<your-platform-release-name>`, find it by running:

`$ helm ls -n <your-platform-namespace>`
````

File: enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md [hunk 33]
```diff
@@ -98,4 +99,45 @@ helm3 upgrade --namespace $NAMESPACE \
...
+To update any of your existing settings, modify them directly in `config.yaml`. To update a setting you haven't already specified, copy the corresponding key-value pair from the [default configuration file](https://github.com/astronomer/docs/blob/main/enterprise/v0.16/reference/default.yaml) into your `config.yaml` file and modify the value from there.
+
+When you have finished updating the key-value pairs, ensure that they have the same relative order and indentation as they do in the default configuration file. If they don't, your changes might not be properly applied.
+
+#### 3. Push the changes to your Astronomer platform.
+
+Save `config.yaml`, then run the following command:
+
+```sh
+$ helm -n <your-namespace> upgrade <your-platform-release-name> astronomer/astronomer --version=<your-platform-version> -f config.yaml
```
paolaperaza, line 134: I'd specify `<your-platform-namespace>` here - individual deployments have namespaces and then the platform has a namespace, so worth specifying I think

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Since this comment thread is outdated I'm resolving and following up in 0.16 file!

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: Made a separate comment on this, but should we be adding `Step 1:` like we do in Cloud Quickstart? Or do we not want to standardize that?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: @jwitz Following up on a different comment, but I'm still seeing this command be different than the ones we have in other docs. Should it be:

```
$ helm upgrade <your-platform-release-name> -f config.yaml --version=<your-current-platform-version> astronomer/astronomer -n <your-platform-namespace>
```

paolaperaza (changes requested):

enterprise/v0.15/06_manage-astronomer/05_upgrade-astronomer.md
paolaperaza: @jwitz I think this is a good tid-bit - since this is the "source of truth" on how to run these commands, I think we should add these examples so people have the right values. Can you add:

- Guidance on how to find namespace
- Guidance on how to find Astronomer Release Name

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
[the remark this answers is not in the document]
paolaperaza: @jwitz Looks like you went down the route of adding brackets with fill-in variables since it's very common for customers to have platform namespace + platform release name _not_ as `astronomer`

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: I went with more formal steps for the Quickstart because people just getting started might appreciate something a little more detailed and clean for an initial install. Also, because of the various substeps for each step, I thought having a more distinguished structure for the top-level steps would make it easy to follow. In the future, a potential standard we could establish is writing out `Step` if it's at the h2 level and writing only the number at the h3 level. Let me know what you think about that!

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Good catch, updated headings throughout the document. Got a bit of tunnel vision with Ian's comments, so I hadn't updated this doc to match the style guidelines we've since put in place. Updating that now!

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Sounds good! Updated in latest commit.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Made both suggestions for finding values into a single tip box, since it's not necessarily essential to the actual step we're describing. Before, it looked like you should have run those commands **after** running the main command.

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Going forward, should we have a style for commands? I could see having commands + subcommands coming first, followed by all flags in alphabetical order. I tried this for the two commands in this section of the doc, let me know if you think it works.

jwitz:
Thanks for the comments @paolaperaza ! All changes based on your comments are in commit e0eb87e485e16ae9481e442e35bd0f0c2301b146. I also responded to a few comments with some potential style guide additions to consider.

paolaperaza (commented):
@jwitz I just added a few final comments! Per our conversation in Slack, a few things:

- Let's split out everything that follows "Configure Platform Settings Between Upgrades" to a new, standalone doc in the "Manage Astronomer" section of the Enterprise menu. It can be called something like, "Apply a Platform Configure Change." This doc should be present only in the `0.16` and `/next` folders.
- Let's standardize adding `your` to reference of `config.yaml`
- Check out my comments on the helm commands at the end of this doc

I'll loop back so we can finally finish this off! Thanks.

enterprise/next/03_install/01_aws/02_install-aws-terraform.md
```diff
@@ -192,4 +192,4 @@ The kubeconfig file along with other secrets such as the TLS certificate are bac
 
 ## Configuring the Platform
 
-Astronomer is deployed on Kubernetes with the [Helm](https://helm.sh) package manager. After completing your installation with Terraform, we recommend updating your configuration through Helm as described in the [Update Your Platform Configuration](https://www.astronomer.io/docs/enterprise/v0.16/manage-astronomer/upgrade-astronomer/update-your-astronomer-configuration-without-upgrading) topic.
+We recommend using Terraform only for your initial installation of Astronomer. Afterwards, all post-install upgrades and configurations to Astronomer should be completed using the [Helm](https://helm.sh) package manager. For more information on how to configure Astronomer after installation, read [Configure Your Astronomer Platform](/docs/enterprise/v0.17/manage-astronomer/configure-platform).
```
jwitz, line 195: For Next docs, should we assume that relative links will increment the file structure like so? Wasn't sure whether to use 0.17 in the filepath or `next`.

jwitz:
@paolaperaza created the new docs! Since the doc is only a single process with steps, I used the same step naming conventions we used for the Quickstart guide. Added one small question about using relative filepaths for `Next` Enterprise docs, but other than that this should be good to go.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -47,12 +49,12 @@ First, ensure you have a copy of the `config.yaml` file of your platform namespa
...
 ```
 
 Review this configuration and delete the line `"USER-SUPPLIED VALUES:"` if you see it.
 
-### Verify your current Platform Version
+### Verify your Current Platform Version
```
paolaperaza, line 57: @jwitz It looks like this `/next` file for "Upgrade Astronomer" doesn't have the same format as 0.16  - did we want to apply the steps method?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -40,27 +41,27 @@ Read below for specific guidelines.
 >
 > Now, Airflow Chart changes only take effect when another restart event is triggered by a user (e.g. a code push, Environment Variable change, resource or executor adjustment, etc).
 
-### Ensure you have a copy of your Astronomer `config.yaml`
+### 1. Ensure You Have a Copy of Your Astronomer config.yaml File
```
paolaperaza, line 44: @jwitz Did we decide on bringing this down to not capitalized title since it's an H3?

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -98,4 +99,4 @@ helm3 upgrade --namespace $NAMESPACE \
 
 If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/enterprise/v0.16/resources/release-notes) from an earlier minor version, submit a request to [Astronomer Support](https://support.astronomer.io).
 
-We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
+We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Winter 2020.
```
paolaperaza, line 102:
```suggestion
We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Winter 2021.
```

paolaperaza (changes requested):

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -13,6 +13,8 @@ To help you maintain Astronomer up-to-date, the guide below will walk you throug
 - Astronomer Platform Versioning
 - How to upgrade to a patch version on Astronomer
 - How to upgrade to a minor version on Astronomer
+- How to apply a platform configuration change
```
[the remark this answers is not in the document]
jwitz: Yep!

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
```diff
@@ -99,5 +101,4 @@ helm3 upgrade --namespace $NAMESPACE \
 
 If you're looking to upgrade to Astronomer Enterprise [v0.16 (latest)](/docs/enterprise/stable/resources/release-notes) from an earlier minor version, submit a request to [Astronomer Support](https://support.astronomer.io).
 
-We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Fall 2020.
-
+We're working on a more robust and reliable upgrade process for our next Astronomer Enterprise "Long-term Support" quarterly release scheduled for Winter 2020.
```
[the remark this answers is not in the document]
jwitz: Fixing for both this doc and in 0.16

enterprise/next/06_manage-astronomer/06_configure-platform.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Configure Your Astronomer Platform"
```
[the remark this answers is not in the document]
jwitz: Good call, I like that reasoning. Updated!

enterprise/next/06_manage-astronomer/06_configure-platform.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Step 3: Push Changes to Your Astronomer Platform
+
+Save your `config.yaml` file, then run the following command:
+
+```sh
+$ helm upgrade <your-platform-release-name> astronomer/astronomer -f config.yaml -n <your-platform-namespace> --version=<your-platform-version>
+```
+
+> **Tip:** The value for `<your-platform-namespace>` can be found in your list of active namespaces. To show this list, run:
```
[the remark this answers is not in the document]
jwitz: Agreed it'd be nice to have CSS styling differentiating the two. I always considered a Note to be something that can cause a reader reconsider or alter their setup actions. I styled this as a tip since it's not really necessary to read if you already know what these names are, and it's not really specific to the setup either. For something this size, I like categorizing it as a tip so readers don't read through assuming it's essential for setup.

enterprise/next/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Yep, got focused on the new doc and forgot to update the Upgrade doc with new standards as well. Good catch

enterprise/v0.16/06_manage-astronomer/05_upgrade-astronomer.md
jwitz: Yep, got focused on the new doc and forgot to update the Upgrade doc with new standards as well. Good catch

jwitz:
Thanks @paolaperaza, all version info and titling should be accurate now!

enterprise/next/06_manage-astronomer/06_configure-platform.md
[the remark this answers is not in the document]
paolaperaza: Good call - I'm in for that!

paolaperaza (approved):
Alright, @jwitz ! I think we are finally done with this one. Huge thanks for the thoughtfulness in our back-and-forths - I'm confident in this last iteration.

Let's ship!

PR #91, opened 2020-12-03
Replaced every instance of 'open-source' with 'open source'

jwitz:
Closes https://github.com/astronomer/issues/issues/2296. Per discussion on Slack, agreement was reached that it's OK to refrain from hyphenation across all uses of open source. This is both how it's done in IBM's developer style guide, the AP style guide, and Google's open source documents.


paolaperaza (approved):
Noice! Looks good @jwitz - thanks for the cleanup.

PR #92, opened 2020-12-04
Add/update GQL to code fence blocks

ryanahamilton:
Resolves astronomer/astro-site#194

I _think_ I got em all.


paolaperaza (approved):
Awesome - thanks @ryanahamilton. I tried looking through other files to check if we missed any others but didn't find at first 2 glances.

cc: @jwitz You may already have this in mind, but let's add "syntax highlighting" guidelines to our docs style guide and/or README for future reference!

samblackk:
@jwitz We're using `prismjs` for syntax highlighting in our code blocks. Here is a list of supported `tags` that can be used: 
https://prismjs.com/#supported-languages

PR #98, opened 2020-12-08
Issue template creation

jwitz:
Added a new issue template to be used for all doc issues going forward. This is the same one that was in the private repo, save for the updated label names.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_template.md
+++ b/.github/ISSUE_TEMPLATE/issue_template.md
@@ -36,6 +36,6 @@ _What is the purpose of the document? Select all that apply:_
   
 #### _Describe your suggestion for the change. In addition, include links to any relevant existing documentation, support tickets, messages, or external content:_
   
-### Context
-Add any context about the problem here.
+### Other Context
+Add any additional context about the problem here.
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/issue_template.md
```diff
@@ -0,0 +1,41 @@
...
+#### _Describe what should be included in the document, making sure to outline what users should be able to achieve with the doc. In addition, include links to any relevant existing documentation, support tickets, messages, or external content:_
+  
+#### _Are you the right person to review a draft of the document for technical accuracy? If not, who is?_ 
+  
+### For fixes, complete the following sections:
+  
+#### _Link to affected document:_
+  
+#### _How severe is the problem?_
+- [ ] Important to fix: The document has incorrect or missing information that is preventing customers from being successful (Add the `error/itf` label to this issue).
```
paolaperaza, line 33:
```suggestion
- [ ] Critical: The document has incorrect or missing information that is preventing customers from being successful (Add the `error/itf` label to this issue).
```

I like the term "critical", but don't feel that strongly. In any case, I think the important thing here is that these "categories" of prioritization match the labels in this repo - i.e. if the person creates a new issue and checks off "Useful" - they should also add a "useful" label to the issue.

paolaperaza (approved):
This is a good start - thanks @jwitz!

paolaperaza:
Actually, @jwitz I'm thinking about this more - since all issues created in this repo are going to be related to documentation - let's maybe split up templates to these different categories?

What if we had a template for:
- New Doc Request
- Enhancement to Existing Doc
- Typo/Error

Just a thought - let me know what you think.

Issue #109, opened 2020-12-10
Need a version of our 'Support' guide in Enterprise docs.


jwitz:
## What kind of documentation change are you requesting?

* **New document, topic, or other piece of content**: Add the `newcontent` label to this issue. 
* **Significant change to an existing document**: Add the `enhancement` label to this issue. 
* **Fix to an existing document**: See below for what labels to add
  
#### _How severe is the problem?_
- [X ] Critical: The document has incorrect or missing information that is preventing customers from being successful (Add the `error/itf` label to this issue).
- [ ] FYI: The document could benefit from a change, but there's nothing that's currently preventing customers from being successful without it (Add the `error/useful` label to this issue).
- [ ] Typo: There's a noticeable grammar, spelling, or formatting issue in a document (Add the `error/typo` label to this issue). 
  
### Other Context
Add any additional context about the problem here.

PR #110, opened 2020-12-10
Added our Support page to Enterprise docsets

jwitz:
This PR closes https://github.com/astronomer/docs/issues/109 . 

I had to also rename some of the files around Support to ensure that each Enterprise docset has the same order of docs under the `Resources` tab. @paolaperaza Is the way I did this fine for the website? I think the numbers before files don't show up in URLs, so changing them shouldn't break any links we have.


paolaperaza:
Awesome, thanks Jake! This looks great - I took a look at those file name changes and they look good. Thank ya!

paolaperaza (approved):

PR #124, opened 2020-12-15
correcting the experimental api url

krisdock:
we omitted "/airflow/" from the url on two docs


jwitz (approved):
Looks good, just added these changes to the 'next' Enterprise docset. Thanks @krisdock !

PR #177, opened 2020-12-30
New cron job for checking broken links

jwitz:
Fixes https://github.com/astronomer/issues/issues/2213#issuecomment-752254937

Created a GitHub action which checks our site for broken links every Sunday at 8am GMT. The results will appear in the `Actions` tab on GitHub.

### DIFF
--- a/linkchecker.yml
+++ b/linkchecker.yml
@@ -1,0 +1,14 @@
+on:
+  schedule:
+    - cron: '0 8 1 * *'  # Runs on the 1st of every month at 8AM GMT
+
+name: Check markdown links
+jobs:
+  my-broken-link-checker:
+    name: Check broken links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check
+        uses: ruzickap/action-my-broken-link-checker@v1
+        with:
+          url: https://astronomer.io
--- a/linkchecker.yml
+++ b/linkchecker.yml
@@ -2,10 +2,10 @@ on:
   schedule:
     - cron: '0 8 * * 0'  # Runs every Sunday at 8AM GMT
 
-name: Check markdown links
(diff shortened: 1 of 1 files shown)


linkchecker.yml
```diff
@@ -0,0 +1,14 @@
+on:
+  schedule:
+    - cron: '0 8 1 * *'  # Runs on the 1st of every month at 8AM GMT
```
danielhoherd, line 3: It wouldn't hurt to run this more frequently. If we want to run it weekly on Sunday morning:

```suggestion
    - cron: '0 8 * * 0'  # Runs every Sunday at 8AM GMT
```

danielhoherd: https://crontab.guru/#0_8_*_*_0 and `man 5 crontab` in case any of that is confusing.

jwitz: Agreed, weekly runs would be good.

paolaperaza:
@jwitz This looks great! Before we merge, I'd say be more specific on the naming convention as it's not immediately obvious what and where this checks for.

For example, we found that the report for this GH action includes links across the Astronomer Docs, Blog and Guides repo. So the naming convention could be:

> Name: "Check for broken links in Markdown"
> Subname: "Astronomer Docs, Blog, and Guides Repos"

Could also add a comment that this will only check for links that are properly formatted, for example. I'll letchu decide/format!

jwitz:
Sounds good @paolaperaza , just updated. Didn't include a comment about properly formatted links, as that should be fully apparent if you're looking on the website (URLS appearing next to text, etc). We should always catch those formatting errors at the point of publication.

paolaperaza (approved):
Looks good @jwitz ! I'll let you merge whenever you're ready.

PR #180, opened 2021-01-04
Moved GitHub action to the appropriate directory

jwitz:
When creating the file for our broken link checker last week, I missed a key step of creating a specific directory for the action as described here: https://docs.github.com/en/free-pro-team@latest/actions/quickstart

This PR simply deletes the instance of linkchecker.yml from the top level `docs` directory and moves it into the appropriate `.github/workflows` directory.


paolaperaza (approved):
Thanks @jwitz !

PR #188, opened 2021-01-07
Add azure flag to helm config

ianstanton:
- Add azure flag to helm config in installation guide
- This will only apply to LTS 0.23.x and onward
- Related to https://github.com/astronomer/issues/issues/2417


jwitz (approved):
Looks good @ianstanton ! Thanks for adding this - really glad we're setting up for success on 0.23 by getting our docs updated now.

Issue #162, opened 2020-12-21
Add Section to CLI Quickstart Doc (Enterprise) on CLI/Platform Versioning Guidelines


jwitz:
@paolaperaza commented on [Wed Jul 15 2020](https://github.com/astronomer/issues/issues/1544)

As of v0.16, the Astro CLI has improved error handling for Astronomer Enterprise users who may be installing versions of the CLI that are not compatible with the version of the Platform they're running.

We should:

- Document intended functionality for Enterprise users
- Explain any error messages as needed

**Note:** May need content for Cloud users as well.

Related GH Issues:

- https://github.com/astronomer/issues/issues/923
- https://github.com/astronomer/issues/issues/1516

PR #225, opened 2021-01-26
Add Astro CLI <> Platform Versioning guidelines to CLI Quickstart + Versioning Matrix

paolaperaza:
This PR adds guidelines around Astronomer CLI and Platform versioning policy and recommendations, applicable to both Astronomer Cloud and Enterprise. Changes include:

1. Add CLI versions to Enterprise Versioning Compatibility Matrix
2. Add a step to our "Upgrade to Astronomer v0.23" guide that instructs users to upgrade the CLI
3. Add information around Platform <> CLI minor versioning in "CLI Quickstart"

Adding @jwitz as a reviewer here but @vishwas-astro @astrorfox feel free to take a look at this as well and provide feedback on any of the above. Once we're good with the changes here, I'll port them over to Cloud docs + v0.16/next Enterprise versions.

Related Astro CLI PR: https://github.com/astronomer/astro-cli/pull/339

Resolves: https://github.com/astronomer/docs/issues/162

### DIFF
--- a/enterprise/v0.23/02_develop/01_cli-quickstart.md
+++ b/enterprise/v0.23/02_develop/01_cli-quickstart.md
@@ -228,6 +228,8 @@ For every minor version of the Astronomer Platform, a corresponding minor versio
 
 While corresponding minor versions of the Astronomer Platform and the Astronomer CLI are required, subsequent patch versions do not need to match and will not introduce breaking changes. The release of Astronomer v0.23.9, for example, does not mean that v0.23.9 of the Astronomer CLI exists or is required.
 
+### Check Running Astronomer CLI and Platform Versions
+
 To check your working version of the Astronomer CLI (`Astro CLI`) and the Astronomer Platform (`Astro Server Version`), run:
 
 ```
--- a/enterprise/v0.23/02_develop/01_cli-quickstart.md
+++ b/enterprise/v0.23/02_develop/01_cli-quickstart.md
@@ -245,7 +245,7 @@ Astro Server Version: 0.23.9
 Git Commit: 748ca2e9de1e51e9f48f9d85eb8315b023debc2f
 ```
 
-If the minor versions for the two components do not match, you'll receive an error message in your command line with instructions to either upgrade or downgrade the Astronomer CLI accordingly. Here, the listed versions of Astronomer and the Astronomer CLI are compatible because they're both in the v0.23 series. If you're running v0.16.10 of Astronomer and v0.23.2 of the Astronomer CLI, for example, you'll be instructed to downgrade the CLI to the latest in the v0.16 series. If you have access to more than one Astronomer Enterprise installation, `Astro Server Version` will correspond to the `<base-domain>` that you're currently authenticated into.
+Here, the listed versions of Astronomer and the Astronomer CLI are compatible because they're both in the v0.23 series. If the minor versions for the two components do not match, you'll receive an error message in your command line with instructions to either upgrade or downgrade the Astronomer CLI accordingly. If you're running v0.16.10 of Astronomer and v0.23.2 of the Astronomer CLI, for example, you'll be instructed to downgrade the CLI to the latest in the v0.16 series. If you have access to more than one Astronomer Enterprise installation, `Astro Server Version` will correspond to the `<base-domain>` that you're currently authenticated into.
 
 For more information on Astronomer and Astronomer CLI releases, refer to:
 
--- a/enterprise/v0.23/02_develop/01_cli-quickstart.md
+++ b/enterprise/v0.23/02_develop/01_cli-quickstart.md
@@ -224,11 +224,11 @@ $ astro dev start
 
 ## Astronomer CLI and Platform Versioning
 
-For every minor version of the Astronomer Platform, a corresponding minor version of the Astronomer CLI is made available. To ensure reliability, upgrade to the corresponding minor version of the Astronomer CLI. If you're on Astronomer v0.23+, for example, Astronomer CLI v0.23+ is required.
+For every minor version of Astronomer, a corresponding minor version of the Astronomer CLI is made available. To ensure that you can continue to develop locally and deploy successfully, you should always upgrade to the corresponding minor version of the Astronomer CLI when you upgrade to a new minor version of Astronomer. If you're on Astronomer v0.23+, for example, Astronomer CLI v0.23+ is required.
 
-While corresponding minor versions of the Astronomer Platform and the Astronomer CLI are required, subsequent patch versions do not need to match and will not introduce breaking changes. The release of Astronomer v0.23.9, for example, does not mean that v0.23.9 of the Astronomer CLI exists or is required.
+While upgrading to a new minor version of Astronomer requires upgrading the Astronomer CLI, subsequent patch versions will remain compatible. For instance, consider a system where Astronomer is on v0.23.9 and the Astronomer CLI is on v0.23.2. While we encourage users to always run the latest available version of all components, these patch versions of Astronomer and the Astronomer CLI remain compatible because they're both in the v0.23 series.
 
-### Check Running Astronomer CLI and Platform Versions
+### Check Running Versions of Astronomer and the Astronomer CLI
 
 To check your working versions of Astronomer (`Astro Server Version`) and the Astronomer CLI (`Astro CLI`), run:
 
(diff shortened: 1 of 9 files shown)


astrorfox:
Looks good to me for the current posture...eventually will want to decouple these further (i.e. not every release of Enterprise Platform will require a version increment of the CLI)

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
 $ astro dev start
 ```
 
-## Additional Resources
+## Astronomer CLI and Platform Versioning
+
+For every minor version of the Astronomer Platform, a corresponding minor version of the Astronomer CLI is made available. To ensure reliability, upgrade to the corresponding minor version of the Astronomer CLI. If you're on Astronomer v0.23+, for example, Astronomer CLI v0.23+ is required.
```
jwitz, line 227:
```suggestion
For every minor version of the Astronomer platform, a corresponding minor version of the Astronomer CLI is made available. To stay up to date, upgrade to the corresponding minor version of the Astronomer CLI whenever you upgrade your Astronomer platform. If you're on Astronomer v0.23+, for example, Astronomer CLI v0.23+ is required.
```

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
...
+
+While corresponding minor versions of the Astronomer Platform and the Astronomer CLI are required, subsequent patch versions do not need to match and will not introduce breaking changes. The release of Astronomer v0.23.9, for example, does not mean that v0.23.9 of the Astronomer CLI exists or is required.
```
jwitz, line 229:
```suggestion
While upgrading to a minor version of Astronomer requires upgrading the Astronomer CLI, subsequent patch versions do not need to match and will not introduce breaking changes. For instance, upgrading from Astronomer v0.23.0 to v0.23.9 would not require upgrading the Astronomer CLI.
```

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
...
+
+### Check Running Astronomer CLI and Platform Versions
+
+To check your working version of the Astronomer CLI (`Astro CLI`) and the Astronomer Platform (`Astro Server Version`), run:
+
+```
```
jwitz, line 235:
```suggestion
```sh
```

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
...
-## Additional Resources
+## Astronomer CLI and Platform Versioning
+
+For every minor version of the Astronomer Platform, a corresponding minor version of the Astronomer CLI is made available. To ensure reliability, upgrade to the corresponding minor version of the Astronomer CLI. If you're on Astronomer v0.23+, for example, Astronomer CLI v0.23+ is required.
+
+While corresponding minor versions of the Astronomer Platform and the Astronomer CLI are required, subsequent patch versions do not need to match and will not introduce breaking changes. The release of Astronomer v0.23.9, for example, does not mean that v0.23.9 of the Astronomer CLI exists or is required.
+
+### Check Running Astronomer CLI and Platform Versions
+
+To check your working version of the Astronomer CLI (`Astro CLI`) and the Astronomer Platform (`Astro Server Version`), run:
```
jwitz, line 233:
```suggestion
To check your working versions of Astronomer (`Astro Server Version`) and the Astronomer CLI (`Astro CLI`), run:
```

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
...
+This command will output something like the following:
+
+```sh
+$ astro version
+Astro CLI Version: 0.23.2
+Astro Server Version: 0.23.9
+Git Commit: 748ca2e9de1e51e9f48f9d85eb8315b023debc2f
+```
+
+If the minor versions between the two don't match, you'll receive an error message in your command line with instructions to either upgrade or downgrade. If you have access to more than one Astronomer Enterprise installation, `Astro Server Version` will correspond to the `<base-domain>` that you're currently authenticated into.
```
jwitz, line 248:
```suggestion
If the minor versions between the two components don't match, you'll receive an error message in your command line with instructions to either upgrade or downgrade. If you have access to more than one Astronomer Enterprise installation, `Astro Server Version` will correspond to the `<base-domain>` that you're currently authenticated into.
```

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
...
+
+```
+$ astro version
+```
+
+This command will output something like the following:
+
+```sh
+$ astro version
+Astro CLI Version: 0.23.2
```
jwitz, line 243:
```suggestion
Astro CLI Version: 0.23.9
```
Since we don't show the error message here, should we have this example be something that wouldn't give the error?

enterprise/v0.23/02_develop/01_cli-quickstart.md
```diff
@@ -220,20 +222,43 @@ Then, restart the Docker containers by running:
...
+Astro Server Version: 0.23.9
+Git Commit: 748ca2e9de1e51e9f48f9d85eb8315b023debc2f
+```
+
+If the minor versions between the two don't match, you'll receive an error message in your command line with instructions to either upgrade or downgrade. If you have access to more than one Astronomer Enterprise installation, `Astro Server Version` will correspond to the `<base-domain>` that you're currently authenticated into.
 
-For more information on the Astronomer CLI, feel free to reference:
+For more information on Astronomer Platform and CLI releases, refer to:
```
jwitz, line 250:
```suggestion
For more information on Astronomer and Astronomer CLI releases, refer to:
```

jwitz (changes requested):
Mostly minor changes and suggestions @paolaperaza , but I was more curious about the usage of "Astronomer Platform" here. It sounds like a product offering, yet we say "Astronomer v0.23" instead of "Astronomer Platform v0.23". 

Based on the recent Slack poll, it looks like most of our team refers to individual instances of Astronomer as "installations".  Should we use that term when referring to a specific instance of Astronomer, and then use "Astronomer Enterprise" when referring to the actual product offering?

enterprise/v0.23/02_develop/01_cli-quickstart.md
paolaperaza: What about:

```suggestion
While upgrading to a new minor version of Astronomer requires to a corresponding minor version of the Astronomer CLI, subsequent patch versions of each will be released on an ad-hoc basis and not necessarily map to each other directly. For instance, v0.23.2 may be the latest available version of the Astronomer CLI, while v0.23.9 is the latest available version of Astronomer. While we encourage users to always run the latest version of all components, patch versions of both Astronomer and the Astronomer CLI within the v0.23 series will always be compatible with one another and will not introduce breaking changes.
```

enterprise/v0.23/02_develop/01_cli-quickstart.md
paolaperaza: @jwitz This wouldn't show an error! The whole point is that you have to be on `0.23.x` of both, but the two won't map 1:1.

The latest version of the Astronomer CLI = v0.23.2
The latest version of Astronomer = v0.23.9

They are compatible with one another - v0.23.9 of the Astro CLI does not exist and will never necessarily exist

paolaperaza (commented):
@jwitz Thanks for reviewing! Committed some of your suggestions and responded with some comments on others - let me know what you think.

enterprise/v0.23/02_develop/01_cli-quickstart.md
jwitz: It sounds like having an out-of-date CLI can cause breakages based on the potential error message mentioned below. I wanted to use firmer language because of that, but seeing as this is in the quickstart I can see why we'd want to be a little more welcoming. Also doing this won't necessarily "ensure" a frictionless experience on its own, so I'd move away from that language.

What about: "To deploy successfully, you should always upgrade to the corresponding minor version of the Astronomer CLI when you upgrade to a new minor version of Astronomer." Any variation on that is cool with me!

enterprise/v0.23/02_develop/01_cli-quickstart.md
jwitz: Cleaning this up a bit:
```suggestion
While upgrading to a new minor version of Astronomer requires upgrading the Astronomer CLI, subsequent patch versions of each product will remain compatible. For instance, consider a system where Astronomer is on v0.23.9 and the Astronomer CLI is on v0.23.2. While we encourage users to always run the latest version of all components, patch versions of both Astronomer and the Astronomer CLI within the v0.23 series will always be compatible with each other such that these versions would be compatible with each other. 
```
I like the use of "series" here!

enterprise/v0.23/02_develop/01_cli-quickstart.md
jwitz: Ah yep my bad! Confused minor and patch here.

jwitz (approved):
Looks good @paolaperaza ! Feel free to implement whatever part of my secondary suggestions you want, but I think your suggestions also look good.

PR #241, opened 2021-02-06
Switch to docker image that has docker image installed.

danielhoherd:
Fixes https://github.com/astronomer/issues/issues/2398 by switching from our docker image to one that is available on most CircleCI nodes, which has docker installed.


danielhoherd:
Ugh, duplicate of https://github.com/astronomer/docs/pull/138/files

Issue #240, opened 2021-02-05
Gitlab users shouldn't use dind in their CI/CD pipelines


jwitz:
**Affected Document:** https://www.astronomer.io/docs/enterprise/v0.23/deploy/ci-cd#gitlab
  
**Severity:** 

- [X] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [ ] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [ ] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**  We shouldn't be recommending Gitlab users to use `dind` in their scripts. According to @danielhoherd, it's possible to use both Kaniko, as well as a bare metal setup. Because we haven't yet tested Kaniko, I think the best option for now is to recommend a bare metal setup. 

**Additional Notes:** I'm not exactly sure how to modify the script to reflect the above change. Assuming for now that I simply need to update the `services: docker:dind` to `services:docker`, but hopefully a tech reviewer will clarify that.

PR #245, opened 2021-02-11
Removed our recommendation to use docker:dind in CI/CD GitLab configs

jwitz:
Resolves https://github.com/astronomer/docs/issues/240

Since we haven't tested Kaniko yet, and @danielhoherd said that we can run GitLab on bare metal, I simply removed the `services` configuration. Wasn't sure whether to do that or remove only the `dind` part of the service. 

Adding @danielhoherd  and @wolfier  for review, since they were in the original thread where this issue was reported. If this looks good, I'll push the change to other versions of this doc.

### DIFF
--- a/enterprise/v0.23/04_deploy/06_ci-cd.md
+++ b/enterprise/v0.23/04_deploy/06_ci-cd.md
@@ -305,8 +305,6 @@ pipelines:
 astro_deploy:
   stage: deploy
   image: docker:latest
-  services:
-    - docker:dind
   script:
     - echo "Building container.."
     - docker build -t registry.gcp0001.us-east4.astronomer.io/infrared-photon-7780/airflow:CI-$CI_PIPELINE_IID .
@@ -376,7 +374,7 @@ trigger:
 
 pool:
   vmImage: 'ubuntu-latest'
- 
+
 variables:
 - group: Variable-Group
 - group: Key-Vault-Group


danielhoherd:
I don't think this will work. This will just make the `docker build` step fail.

jwitz:
> I don't think this will work. This will just make the `docker build` step fail.

@danielhoherd How would we restructure this configuration so that we're using neither Kaniko nor `docker:dind`?

danielhoherd:
@jwitz Without seeing the original ticket requirements for this section I'm not sure how to best improve this. The Gitlab docs are pretty extensive and cover this topic. <https://docs.gitlab.com/ee/ci/docker/using_docker_build.html>

One of the complicated things with Gitlab is that it has so many options on what environment CI runs in. At the most broad end are the [executors](https://docs.gitlab.com/runner/executors/), but then those executors can in various ways so that complicates things further.

If we are going under the assumption that CI jobs are running in the [Gitlab shared runners](https://docs.gitlab.com/ee/user/gitlab_com/index.html#shared-runners), then the change you made might actually be the right change, because it looks like those are virtual machines running in [docker machine](https://github.com/docker/machine#docker-machine), not docker containers. One way to test your proposed changes is to make a repo on gitlab and try it out. We just need to test the `docker build` part, which should be pretty easy using our docs as a starting point.

jwitz:
@danielhoherd @wolfier I gave this a shot myself, and I think I'm a little too inexperienced with GitLab to know whether my config failed because of user error vs. functional error. Can one of you re-confirm that this setup is minimally viable for users implementing CI/CD using GitLab?

danielhoherd:
@jwitz i just tested this and the changes you made break the gitlab setup. the pre-change config works. this is exactly what i used:

```
---
image: docker:latest

astro_deploy:
  services:
    - docker:dind
  stage: build
  script:
    - echo "Building container.."
    - docker build -t airflow:CI-$CI_PIPELINE_IID .
  only:
    - main
    - master
```

I thought `CI_PIPELINE_IID` was a typo, but apparently it's not because it worked, and I see it listed in the docs.

If you want to zoom and go through this it wouldn't take long.

jwitz:
Thanks for running this and meeting with me to discuss the issue further @danielhoherd. Short-term, I think we can nix this PR given that `dind` still works as an MVP when building an image on GitLab. Long term, I think there's still confusion around CI/CD and what we should documenting around it.

A recent PR added a high level example CI/CD workflow to this doc, but I think we could go further with being explicit in what's happening within these pipelines and how they connect to Airflow. Perhaps we could split up the CI/CD doc into two docs: one explaining CI/CD on Astronomer, and a reference guide with our collection of templates.

PR #272, opened 2021-03-01
Document how to use AWS Cognito user pools as an identity provider

pgagnon:


jwitz (approved):
This looks great @pgagnon ! All I did was add some minor wording updates and pushed this content to other doc versions, so I think it's good to merge. Thank you for contributing this.

PR #288, opened 2021-03-12
Correct Build Image

astrorfox:
A user has pointed out that our build images referenced as ap-airflow is incorrect and should be ap-build. Without Docker installed in ap-airflow, it will not successfully build.


jwitz (approved):
Awesome, thank you for this @astrorfox ! I pushed these changes to the latest Enterprise docs as well. With that, I think this is ready to merge.

PR #295, opened 2021-03-23
Created a reference guide for AC's image architecture

jwitz:
This is a starting draft for what is essentially an editorialized version of the Astronomer Core source code. The info here should provide additional context for various components of the software. When we don't have the exact info someone might be looking for, there should always be a link to the most relevant source code/ documentation. 

Most "controversial" call here was to not document the Astronomer FAB security manager. Based on this [Slack thread](https://astronomerteam.slack.com/archives/CGQSYG25V/p1613011191412100), it looks like the security manager isn't actually used in the open source image. It has to be configured, and that info is probably better suited in another document. 

Looking at other image docs online, there's some additional directions we can take with this. Following [Prisma's example](https://v1.prisma.io/docs/1.34/prisma-server/deployment-environments/docker-rty1/), do we want to provide some basic Docker commands for folks to see what's going on in their image? Or is that out of the scope of this document? What else might people be interested in learning about here? Let me know.

### DIFF
--- a/ac/next/image.md
+++ b/ac/next/image.md
@@ -6,25 +6,37 @@ description: "A reference guide of every minimum package and process required to
 
 ## Overview
 
+The Astronomer Core image for Apache Airflow extends the community-developed Airflow image in a way that makes running Airfow more secure, reliable, and extensible.
+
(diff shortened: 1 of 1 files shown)


paolaperaza (approved):
@jwitz Per our chat in Slack, let's get this merged since we'll no longer be focusing on this docs suite, and merging won't publish this publicly either way. I am sure we will come back to this in the future!

PR #317, opened 2021-04-02
Applied changes from previous PR to other installs

jwitz:
Same changes as in https://github.com/astronomer/docs/pull/304/ , but for other install guides as well.


paolaperaza (approved):
TY @jwitz !

Issue #318, opened 2021-04-06
Proposal Cleanup after LTS Upgrade Script


stevenjs1:
**Affected Document:**
  https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/upgrade-to-0-23
**Severity:**

- [ ] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [X] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [ ] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**
It is best practice in Kubernetes to clean up resources after an upgrade. 
We should add a step to remove all resources after the upgrade. 
kubectl delete -f https://raw.githubusercontent.com/astronomer/astronomer/master/bin/migration-scripts/lts-to-lts/0.16-to-0.23/manifests/upgrade-0.16-to-0.23.yaml 

**Additional Notes:**

Issue #320, opened 2021-04-09
Azure AD integration requires an update to redirect base from 0.16 to 0.23


rbankston:
**Affected Document:**
  https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/upgrade-to-0-23

**Severity:**

- [x] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [ ] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [ ] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**
Under the before you begin section call out the change to Redirect url need to be update with "/" at the end in the Azure AD -> App Registration [auth instructions for 0.23](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/integrate-auth-system)

**Additional Notes:**
https://astronomer.zendesk.com/agent/tickets/4239

PR #323, opened 2021-04-12
Improved the Upgrade to 0.23 doc on multiple fronts

jwitz:
Closes https://github.com/astronomer/docs/issues/318, https://github.com/astronomer/docs/issues/320, and https://github.com/astronomer/docs/issues/319.

Multiple improvements to this doc: two new pre-install considerations, and one post-install step.

Question for @stevenjs1 : Is the K8s resource cleaning step in the right place?  I assume people will want to run this only after they've confirmed that the installation was successful, but let me know if that's not the case. 

Question for @rbankston: Are `deployment.serviceaccounts.get` and `workspace.users.get` new values? If so, how would they be able to set them before the upgrade? If you have a link to the breaking change in GH that resulted in this ticket, that would be super helpful.


rbankston (approved):
They look awesome thank you @jwitz

rbankston:
> Question for @rbankston: Are `deployment.serviceaccounts.get` and `workspace.users.get` new values? I

They aren't new values but new requirements for the UI to render a page.

enterprise/v0.23/06_manage-astronomer/06_upgrade-to-0-23.md
```diff
@@ -10,9 +10,12 @@ As of Astronomer v0.16, Astronomer releases will be made generally available to
...
+A few important notes before you start:
+
 - You must be on Astronomer Enterprise v0.16.x in order to upgrade to Astronomer 0.23+. If you are running a version of Astronomer that's lower than v0.16, submit a request to [Astronomer Support](https://support.astronomer.io) and our team will help you define an alternate upgrade path.
 - The guidelines below only apply to users who are upgrading to the Astronomer v0.23 series for the first time. Once you've completed the upgrade to any v0.23 version, you'll be free to upgrade to subsequent v0.23.x patch versions as they are released by our team. For instructions, read [Upgrade to a Patch Version](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/upgrade-astronomer-patch).
+- If you use Azure AD as your auth system for Astronomer, ensure that your Redirect URI is formatted as `https://houston.BASEDOMAIN/v1/oauth/redirect/`. In v0.16 and prior versions, this URI was formatted as `https://houston.BASEDOMAIN/v1/oauth/redirect`. For information on configuring this value, read [Integrate an Auth System](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/integrate-auth-system#azure-ad).
```
paolaperaza, line 17:
```suggestion
- If you use Azure AD as your auth system for Astronomer, ensure that your Redirect URI is formatted as `https://houston.BASEDOMAIN/v1/oauth/redirect/`. In v0.16 and prior versions, this URI was formatted as `https://houston.BASEDOMAIN/v1/oauth/redirect`, without the extra `/`. For information on configuring this value, read [Integrate an Auth System](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/integrate-auth-system#azure-ad).
```

paolaperaza: Adding this only bc it took me ~45 seconds to comb though these to identify the difference in those URLs - can't hurt to point it out?

enterprise/v0.23/06_manage-astronomer/06_upgrade-to-0-23.md
```diff
@@ -10,9 +10,12 @@ As of Astronomer v0.16, Astronomer releases will be made generally available to
...
+- If you altered the default permissions for user roles as described in [Customize Permissions](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/manage-platform-users#customize-permissions), you need to set both `deployment.serviceAccounts.get` and `workspace.users.get` to `true` in your `config.yaml` file to enable users of this custom role to access Airflow Deployments on Astronomer. To update these values in your system, follow the steps in [Apply a Config Change](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/apply-platform-config).
```
paolaperaza, line 18:
```suggestion
- If you altered the default permissions for user roles as described in [Customize Permissions](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/manage-platform-users#customize-permissions), you need to set both `deployment.serviceAccounts.get` and `workspace.users.get` to `true` in your `config.yaml` file to make sure that all users of a custom role can continue to access Airflow Deployments via the Astronomer UI. To update these values in your system, follow the steps in [Apply a Config Change](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/apply-platform-config).
```

Suggested slight change in wording since it's not so much "enabling" as much as it is "so it doesn't break" (since it working is expected). If they don't do this, the Deployment view in the Astro UI just won't load (per the ZD ticket).

paolaperaza (approved):
Looks great @jwitz! Approving with two super minor suggestions.

stevenjs1:
@jwitz Sorry about the late response. I agree, only after they confirmed everything is working as expected should they run that to clear that install meta-data. 
So maybe move that near the end.

Issue #222, opened 2021-01-26
Add Information on updates.astronomer.io Cron Job to "Upgrade Apache Airflow" (Enterprise Only)


paolaperaza:
In contrast with v0.16, Astronomer Enterprise v0.23 does not require that a release be shipped with explicit support for any Astronomer Certified version or set of Docker images. As soon as a version of Astronomer Certified is published to our updates service, all corresponding Astronomer Certified images and tags will become available for selection in the Astronomer UI and CLI within 24 hours.

Content around our updates.astronomer.io cron job is referenced in [v0.23 Release Notes](https://www.astronomer.io/docs/enterprise/v0.23/resources/release-notes), but needs to be added (with more detail) to our [Upgrade Airflow doc](https://www.astronomer.io/docs/enterprise/v0.23/customize-airflow/manage-airflow-versions).

Source code for that cron job here for reference:
- AC: https://github.com/astronomer/houston-api/blob/main/src/scripts/check-airflow-updates/index.js
- Platform: https://github.com/astronomer/houston-api/blob/main/src/scripts/check-platform-updates/index.js

PR #333, opened 2021-04-27
Added info about version decoupling to Airflow Upgrade doc

jwitz:
Closes https://github.com/astronomer/docs/issues/222

Adding @krisdock for review, since I got the `kubectl` command from one of his Slack messages. Please let me know if there's any other info I should include about this feature.


paolaperaza:
Nice, this is a good start @jwitz and something I agree we need to mention explicitly.

Did you end up looking into the Cron job described in the original GH issue? There are times where users want a version of Airflow available _sooner_ than 24 hours, and want to trigger that cron job manually. Or in some cases, there's some issue that causes it not to run.

Curious to get that documented here, even if briefly (and even if the cron job script itself is private).

jwitz:
@paolaperaza Integrated the info I found this morning into the PR. I think this should now cover what you wanted in the original issue.

enterprise/v0.23/05_customize-airflow/01_manage-airflow-versions.md
```diff
@@ -33,6 +33,16 @@ Astronomer Certified offers support for the following versions of Apache Airflow
 - [Airflow 1.10.7](https://github.com/apache/airflow/releases/tag/1.10.7)
 - [Airflow 1.10.5](https://github.com/apache/airflow/releases/tag/1.10.5)
 
+Starting with Astronomer Enterprise v0.23, new versions of Astronomer Certified are automatically pulled into the Astronomer UI and CLI within 24 hours of their publication via a [cron job](http://updates.astronomer.io/astronomer-certified). In other words, you don't have to upgrade Astronomer in order to upgrade Airflow.
```
paolaperaza, line 36:
```suggestion
Starting with Astronomer Enterprise v0.23, new versions of Astronomer Certified are automatically pulled into the Astronomer UI and CLI within 24 hours of their publication via a cron job that pulls from [Astronomer's updater service](http://updates.astronomer.io/astronomer-certified). In other words, you don't have to upgrade Astronomer in order to upgrade Airflow.
```

The cron job isn't quite updates.astronomer.io itself - that's just what it pulls from! Corrected that here, but otherwise looks great. Make sure to add this to v0.25 docs too (in sep PR) 👀

paolaperaza (approved):
Made 1 quick wording correction, but looks great. TY!

krisdock (approved):
lgtm

paolaperaza (approved):
Just did a final look here - looks great @jwitz !

PR #339, opened 2021-05-05
Helm reference

jwitz:


enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### .Values.global.houston.backendConnection
+
+The details of the connection to your backend Houston datastore, defined in key-value pairs. Possible keys are `user`, `pass`, `host`, `port`, and `db`.
+
+| Value Type | Default Value | Valid Values | Importance |
+| ---------- | ------------- | ------------ | ---------- |
+| Key-value pairs    | {}           | N/A     | High     |
+
+#### .Values.global.houston.airflowBackendSecretName
```
jwitz, line 203: Is the "Airflow backend" the Airflow metadata db? Or something else?

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,217 @@
...
+
+The name of the secret for the backend Houston datastore.
+
+| Value Type | Default Value | Valid Values | Importance |
+| ---------- | ------------- | ------------ | ---------- |
+| String    | ~           | Any     | High          |
+
+#### .Values.global.houston.backendSecretName
+
+The details of the connection to your backend Houston datastore, defined in key-value pairs. Possible keys are `user`, `pass`, `host`, `port`, and `db`.
```
jwitz, line 189: Are there any additional configs here? Took these from the inline comments in the chart.

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+|`astronomer.nodeSelector`| Object | `{ }` | The Kubernetes [nodes](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) on which you run the Astronomer platform |
+|`astronomer.affinity`| Object | `{ }` | Specifies [Affinity values](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) for Kubernetes pods running the Astronomer platform. | [Affinity values](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity)   |
+|`astronomer.tolerations`| List | `[ ]` | Specifies whether Kubernetes pods can schedule onto nodes with certain qualities. For more information, see the Kubernetes [taint values documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) |  
+
+
+## Astronomer Images Configurations
+
+The settings Astronomer images to use on your platform are located under the `astronomer.images` key.
```
jwitz, line 40: When someone applies a platform config via Helm Upgrade, do they have to specify the leading `astronomer` in the key? As in, is the real key `images` or `astronomer.images`?

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+
+This reference doc contains information about each configurable value in the [Astronomer Helm chart](https://github.com/astronomer/astronomer/blob/master/charts/astronomer/values.yaml). These values can be set and pushed to your Astronomer installation as described in [Apply a Config Change](/docs/enterprise/v0.25/manage-astronomer/apply-platform-config).
+
+For a list of all Helm charts used to configure Astronomer, including ones for services such as Elasticsearch and Postgres, see the [Charts repository](https://github.com/astronomer/astronomer/tree/master/charts).
+
+## Astronomer Global Configuration
+
+The settings for Astronomer's global Kubernetes resource configuration are located under the `astronomer` key:
+
+```yaml
```
jwitz, line 17: Curious on reviewer opinion here: If we correctly format the values in tables, is there any benefit in including these code fences from directly from GitHub? Which format are admins more likely to look at? I can see the code fences getting out of date frequently compared to the tables.

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+| astronomer.houston.cleanupDeployments.canary | bool | `false` | Determines whether cleanup jobs run only on Deployments marked as `canary` |
+| astronomer.houston.updateCheck.enabled | bool | `true` | | Determines whether to run a cron job that checks for Astronomer updates |
+| astronomer.houston.updateCheck.url | string | `"https://updates.astronomer.io/astronomer-platform"` | The cron job to run for checking Astronomer updates |
+| astronomer.houston.updateCheck.schedule | string | `"0 0 * * *"` | The schedule for running the Astronomer updates check cron job |
+| astronomer.houston.updateAirflowCheck.enabled | bool | `true` | Determines whether to run a cron job that checks for Airflow updates |
+| astronomer.houston.updateAirflowCheck.url | string | `"https://updates.astronomer.io/astronomer-certified"` | The cron job to run for checking Airflow updates |
+| astronomer.houston.updateAirflowCheck.schedule | string | `"0 0 * * *"` | The schedule for running the Airflow updates check cron job |
+| astronomer.houston.airflowReleasesConfig | string | `nil` |  |
```
jwitz, line 348: Not sure what this value does

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+| astronomer.houston.backendSecretName | string | `nil` | The secret name for the Houston datastore |
+| astronomer.houston.backendConnection | object | `{}` | Credentials for the Houston datastore backend connection |
+| astronomer.houston.airflowBackendSecretName | string | `nil` | The secret name for the Airflow backend connection |
+| astronomer.houston.airflowBackendConnection | object | `{}` | Credentials for the Airflow backend connection |
+| astronomer.houston.jwtSigningKeySecretName | string | `nil` | Optional secret name to use to sign JSON web tokens  |
+| astronomer.houston.jwtSigningCertificateSecretName | string | `nil` | Optional secret name to use to sign JSON web tokens |
+| astronomer.houston.env | list | `[]` | Kubernetes environment variables |
+| astronomer.houston.secret | list | `[]` | Kubernetes secret for your environment |
+| astronomer.houston.resources | object | `{}` | Specifies CPU and memory usage limits and requests for the Astronomer UI |
+| astronomer.houston.config | object | `{}` | Additional Houston configs |
```
jwitz, line 327: Could use some more specific details here

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+| astronomer.houston.updateCheck.url | string | `"https://updates.astronomer.io/astronomer-platform"` | The cron job to run for checking Astronomer updates |
+| astronomer.houston.updateCheck.schedule | string | `"0 0 * * *"` | The schedule for running the Astronomer updates check cron job |
+| astronomer.houston.updateAirflowCheck.enabled | bool | `true` | Determines whether to run a cron job that checks for Airflow updates |
+| astronomer.houston.updateAirflowCheck.url | string | `"https://updates.astronomer.io/astronomer-certified"` | The cron job to run for checking Airflow updates |
+| astronomer.houston.updateAirflowCheck.schedule | string | `"0 0 * * *"` | The schedule for running the Airflow updates check cron job |
+| astronomer.houston.airflowReleasesConfig | string | `nil` |  |
+
+## Prisma
+
+Configurations for Prisma are stored under the `astronomer.prisma` key:
```
jwitz, line 352: What specifically are we using Prisma for on Enterprise, I know they offer database connections, but how does that work in the context of Astronomer?

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| prisma.podLabels | object | `{}` | Custom labels for pods running Prisma  |
+| prisma.env | list | `[]` | The default resources used to run Prisma |
+| prisma.resources | object | `{}` |  Specifies CPU and memory usage limits and requests for Prisma |
+
+## Commander
+
+Configurations for Astronomer's Commander service are stored under the `astronomer.commander` key:
```
jwitz, line 377: We don't have Commander documented at all atm, we should include a brief description for what it does here.

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+| astronomer.registry.s3.secretkey | string | `nil` | The encryption key secret for the AWS S3 bucket |
+| astronomer.registry.s3.region | string | `nil` | The region of the AWS S3 bucket |
+| astronomer.registry.s3.bucket | string | `nil` | The ID for the AWS S3 bucket |
+| astronomer.registry.s3.encrypt | bool | `false` |  Determines whether to use encryption for AWS S3|
+| astronomer.registry.s3.keyid | string | `nil` | The encryption key ID for the AWS S3 bucket |
+| astronomer.registry.s3.rootdirectory | string | `nil` | The root directory of the AWS S3 bucket  |
+
+## Install
+
+Configurations for the Astronomer CLI are stored under the `astronomer.install` key:
```
jwitz, line 500: Took a wild guess on this based on some earlier config values. Could definitely use some help with specifics here.

enterprise/v0.23/09_resources/05_config-reference.md
```diff
@@ -0,0 +1,548 @@
...
+| astronomer.registry.persistence.size | string | `"100Gi"` | The size of the persistent storage |
+| astronomer.registry.persistence.storageClassName | string | `nil` | The name reference for all statefulSets, if using a custom persistent storage |
+| astronomer.registry.auth.secretName | string | `nil` | An optional secret name for the Registry certificate authority  |
+| astronomer.registry.auth.connection | string | `nil` | The connection for the Registry certificate authority |
+| astronomer.registry.auth.service | string | `"docker-registry"` | The service to use for generating the Registry certificate |
+| astronomer.registry.auth.issuer | string | `"houston"` | The issuer for the Registry certificate  |
+| astronomer.registry.gcs.enabled | bool | `false` |  Determines whether to use GCS for Astronomer Enterprise |
+| astronomer.registry.gcs.bucket | string | `nil` | The GCS bucket storage to use |
+| astronomer.registry.gcs.useKeyfile | bool | `true` | Determines whether to use a key file for service account credentials |
+| astronomer.registry.gcs.keyfile | string | `"/var/gcs-keyfile/astronomer-gcs-keyfile"` |  |
```
jwitz, line 481: Another value where I'm not quite sure how to describe it

krisdock:
@shmanu017 @rbankston you both have a lot of experience with helm configs. just making sure you see this.

danielhoherd:
While I appreciate the effort and love the goal of this PR, I am worried that it will get out of sync with reality, and will never be 100% accurate across different versions. It would be great if we could supplement this work with an effort to document the config in a way that makes the reference documentation regenerable after making changes to the chart, or at least makes it so that can account for changes and manually document those changes. I think this more holistic approach to documenting the config would make for better design and better documentation, and reduce errors in documentation, especially if we rope the docs folks in early when we are designing stuff.

Many projects document the config within the config, which keeps the docs glommed onto the version being documented. Maybe that would be a good idea, and we can just parse the comments out of the config if we want to format them for the web?

danielhoherd:
Also, it looks like we are using helm-docs to generate some of this, but I don't see that referenced anywhere in code. Just to be sure that we are using tooling where it would be helpful, I want to point out https://github.com/norwoodj/helm-docs

jwitz:
@danielhoherd Totally understand your concerns with keeping this up-to-date, as it's something that we've not traditionally been strong at with docs. 

Automation is one solution here, but the other, which I think we're closer to, is tightening the communication between R&D and docs. The actual changes to this doc should be pretty straightforward whenever the helm configuration changes. But docs is often out of the loop, meaning we update the docs only after the change has been out for a while. If we solve that problem first (formalizing how R&D alerts us about changes with documentation consequences), then I think this doc will actually be pretty easy to maintain.

As for the tooling, I did use the tool you mentioned to generate the tables. We aren't using it to generate the descriptions, though. Until we standardize our inline documentation, we'll have to manually update descriptions for each variable.

danielhoherd:
@jwitz one way to get a better idea about changes to the file would be to add yourself as a CODEOWNER for values.yaml and any other files that are important to you to maintain this doc. I made a PR to do this https://github.com/astronomer/astronomer/pull/1183

jwitz:
In the interest of getting this PR out the door, I'm going to cut down on work by assigning reviewers specific sections of the PR.

@shmanu017 : Can you please review `Astronomer Global Configuration` / `Astronomer Images Configurations` / `Astronomer UI Configuration` ?

@schnie and @rbankston  : Can you please review `Houston Configuration` / `Prisma` / `Commander`?

@wolfier : Can you please review `Registry` / `Install` / `Ports` ?

enterprise/v0.23/09_resources/05_config-reference.md
rbankston: These are any value you want to pass to houston directly.

publicSignups: (true/false)
emailConfirmation: (true/false)
deployments:
         manualReleaseNames: false 
        maxPodAu: 100
        configureDagDeployment: true
        nfsMountDagDeployment: true
        serviceAccountAnnotationKey: eks.amazonaws.com/role-arn

enterprise/v0.23/09_resources/05_config-reference.md
danielhoherd: Yes.

enterprise/v0.23/09_resources/05_config-reference.md
rbankston: Don't believe this matters past 0.16 since prisma was brought into houston. Prisma used to be its own pod.

enterprise/v0.23/09_resources/05_config-reference.md
rbankston: Authorization file for GCS credentials

enterprise/v0.23/09_resources/05_config-reference.md
rbankston: It is the pod that provides a script for installing the astro cli

enterprise/v0.23/09_resources/05_config-reference.md
danielhoherd: This is actually wrong. [user, pass, host, port and db all belong to backendConnection](https://github.com/astronomer/astronomer/blob/51ec05e2e5/charts/astronomer/values.yaml#L77). This is ag good example of why programmatically generated docs would be advantageous.

enterprise/v0.23/09_resources/05_config-reference.md
danielhoherd: If they're in a table then I don't see a good reason to include them here. We could just link to that section of the code on GH.

enterprise/v0.23/09_resources/05_config-reference.md
danielhoherd: Commander is the astronomer helm client. It does all of the helm work related to deployments of airflow into the k8s cluster. I'm sure there's more to it than that, but that's a start!

rbankston: Commander updates the Airflow deployment helm charts.

enterprise/v0.23/09_resources/05_config-reference.md
schnie: Yep - agree with @rbankston!

PR #346, opened 2021-05-10
Fixing the Azure installation for typos and proper default values

park-peter:
The default values in Azure installation document was confusing to the customers due to having many default values that needed to change. Also had `ssl` section of values.yaml in the wrong level.


jwitz (approved):
These changes look great @park-peter ! Thanks for handling this.

Issue #351, opened 2021-05-13
Deployment Viewers are able to view task logs but the docs says they can't


wolfier:
**Affected Document:** [Manage User Permissions on Astronomer Enterprise](https://www.astronomer.io/docs/enterprise/v0.23/manage-astronomer/workspace-permissions)
  
**Severity:**

- [x] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [ ] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [ ] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

Please remove 
```
Cannot view Airflow task logs
```
 ![Screen Shot 2021-05-13 at 11 42 27 AM](https://user-images.githubusercontent.com/5952735/118171648-4ebd3100-b3e0-11eb-8e82-bd054b089bc5.png)

**Additional Notes:**

PR #357, opened 2021-05-17
Fixed Deployment Viewer permission information

jwitz:
Resolves https://github.com/astronomer/docs/issues/351

At some point in v0.23, we wrote that Deployment Viewers cannot view Airflow task logs, which are available in the Airflow UI. I tried to be more clear about what was / wasn't possible for this type of user. By saying "They can only", I avoid having to list out the many things that Deployment Viewers *can't* do.

It seems like we might have mistook task log access for Kibana logging metrics, but I don't think it's even necessary to mention that lack of permission here.


wolfier:
lgtm 🎉

wolfier (approved):

PR #365, opened 2021-05-20
Updated arguments for REST API calls

jwitz:
Resolves https://github.com/astronomer/docs/issues/363

I didn't actually see this argument in the first screenshot referenced in the GitHub issue, so I'm commenting in places where we might additionally want to add the argument.

### DIFF
--- a/cloud/stable/04_customize-airflow/07_airflow_api.md
+++ b/cloud/stable/04_customize-airflow/07_airflow_api.md
@@ -29,11 +29,11 @@ To create a Service Account via the Astronomer UI:
    ![New Service Account](https://assets2.astronomer.io/main/docs/ci-cd/ci-cd-new-service-account.png)
 3. Give your Service Account a **Name**, **User Role**, and **Category** (_Optional_).
    > **Note:** In order for a Service Account to have permission to push code to your Airflow Deployment, it must have either the Editor or Admin role. For more information on Workspace roles, refer to [Roles and Permissions](/docs/cloud/stable/manage-astronomer/workspace-permissions/).
-   
+
    ![Name Service Account](https://assets2.astronomer.io/main/docs/ci-cd/ci-cd-name-service-account.png)
-   
+
 4. Save the API Key that was generated. Depending on your use case, you may want to store this key in an Environment Variable or secret management tool of choice.
-   
+
    ![Service Account](https://assets2.astronomer.io/main/docs/ci-cd/ci-cd-api-key.png)
 
 ### Create a Service Account via the Astronomer CLI
@@ -71,7 +71,7 @@ You can make requests via the method of your choosing. Below, we'll walk through
 
 ### Trigger DAG
 
-If you'd like to externally trigger a DAG run, you can start with a generic cURL command to Airflow's POST endpoint: 
+If you'd like to externally trigger a DAG run, you can start with a generic cURL command to Airflow's POST endpoint:
 
 ```
 POST /airflow/api/experimental/dags/<DAG_ID>/dag_runs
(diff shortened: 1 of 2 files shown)


jwitz:
I don't think GitHub is letting me comment in parts of the code that aren't in the PR, but @ben-astro can you confirm that we don't need to add this argument to the following call?

```
curl -v -X POST
https://<AIRFLOW-DOMAIN>/airflow/api/experimental/dags/<DAG-ID>/dag_runs
-H 'Authorization: <API-Key> ’
-H ‘Cache-Control: no-cache’
-H ‘content-type: application/json’ -d ‘{}’
```

Also, does this change apply to Enterprise v0.23? It looks like in the original GH issue, the issue was replicated in Airflow 2.0.

cloud/stable/04_customize-airflow/07_airflow_api.md
```diff
@@ -183,7 +183,8 @@ Here, your cURL request would look like the following:
 curl -X GET \
 https://<AIRFLOW-DOMAIN>/airflow/api/v1/config \
 -H 'Authorization: <API-Key>' \
--H 'Cache-Control: no-cache'
+-H 'Cache-Control: no-cache' \
+-H "Accept:'application/json'"
```
paolaperaza, line 187:
```suggestion
-H 'Cache-Control: no-cache' \
-H 'Accept:'application/json'
```

The formatting for this argument looks different than the others - is that intentional, or a typo? Made a suggestion that makes it parallel to the rest, but do correct me if I'm wrong.

enterprise/v0.23/05_customize-airflow/08_airflow-api.md
```diff
@@ -183,7 +183,8 @@ Here, your cURL request would look like the following:
 curl -X GET \
 https://<AIRFLOW-DOMAIN>/airflow/api/v1/config \
 -H 'Authorization: <API-Key>' \
--H 'Cache-Control: no-cache'
+-H 'Cache-Control: no-cache' \
+-H "Accept:'application/json'"
```
jwitz, line 187:
```suggestion
-H 'Accept: application/json'
```

enterprise/v0.25/05_customize-airflow/08_airflow-api.md
```diff
@@ -183,7 +183,8 @@ Here, your cURL request would look like the following:
 curl -X GET \
 https://<AIRFLOW-DOMAIN>/airflow/api/v1/config \
 -H 'Authorization: <API-Key>' \
--H 'Cache-Control: no-cache'
+-H 'Cache-Control: no-cache' \
+-H "Accept:'application/json'"
```
jwitz, line 187:
```suggestion
-H 'Accept: application/json'
```

cloud/stable/04_customize-airflow/07_airflow_api.md
jwitz: Ooh good call, we don't need the end slash but there should maybe only be one set of quotes covering the entire param. @ben-astro can you confirm the format here?

enterprise/v0.25/05_customize-airflow/08_airflow-api.md
ben-astro: @jwitz probably better to get @andriisoldatenko to confirm the proper formatting - I would need to do some empirical testing to validate anything I told you.

snolan-amount:
I don't want to distract from the original change here, but in the CURL example being changed I have also found that the explanation of how to specify the Airflow domain is not correct, the docs say
`<AIRFLOW-DOMAIN>`: Use `https://<your-base-domain>/<deployment-release-name>`
In my case I had to use
`<AIRFLOW-DOMAIN>`: Use `https://deployments.<your-base-domain>/<deployment-release-name>`

PR #382, opened 2021-06-07
Update Azure Postgres requirement

park-peter:
From internal testing, it was found that single server Azure PostgreSQL breaks Astronomer. Only Flexible Server is supported.


jwitz (approved):
Thank you for the update @park-peter! Merging now.

PR #391, opened 2021-06-18
fix example and instruction for setting variables in .env file

cmarteepants:
When adding environment variables into a .env file, the syntax is `KEY=VAR`. We accidentally said to use `ENV KEY=VAR`, which is the syntax for adding environment variables to a Dockerfile. 

When users follow the instructions as is, environment variables silently aren't loaded, which leaves a very confused customer :) We should consider adding a check to make sure that environment variables are properly formatted and that the key is uppercase in the future.


jwitz (approved):
Excellent, thank you @cmarteepants for catching this and making the fix!

PR #392, opened 2021-06-21
Potential security doc fix + typo hotfix

jwitz:
Fritz mentioned that the frontmatter of the Security docs had different capitalization from all other docs, which might be what's causing the sidenav to not load. 

Also fixed an egregious typo 🌞


paolaperaza (approved):
Oof, good catches here @jwitz + @fritz-astronomer. Approving!

PR #394, opened 2021-06-21
Replace curly quotes in code blocks with straight quotes

park-peter:
Curly quotes cause syntax error. Replace with straight quotes


paolaperaza (approved):
Yikes - good catch here @park-peter and thanks for making sure your change was applied across files.

Issue #386, opened 2021-06-16
AWS Enterprise NLB - extra documentation / misconfiguration


fritz-astronomer:
**Affected Document:**
https://www.astronomer.io/docs/enterprise/v0.25/install/aws/install-aws-standard
  
**Severity:**
- [ ] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [x] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [ ] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**
A customer (Amount) brought this up. Apparently in a private VPC, if you have an NLB, there's a weird 60s timeout that makes our web interface **very** slow. The fix is to use a classic load balancer - not NLB. 

Our install documentation includes NLB in the default `config.yaml` snippet. We should probably include a note on it, and specifically mention not to use it in this instance. @cmarteepants might have more info around exactly what the issue is

cmarteepants:
Just to clarify the issue - I might have some details wrong - but I encountered this problem when installing a production self-hosted Gitlab instance in EKS.

If the following is true:

- the ec2 nodegroups for EKS are on private subnets
- the vpc does not have a NAT gateway configured. (traffic is first routed to a shared account via transit gateway where they eventually have external access in client's case)

When using AWS NLB with an ingress controller (nginx), if two pods are collocated on the same node and attempt to communicate using an external dns, the source IP and destination IP are - according to nginx - one and the same, and thinks that the packets are malformed and throws an error. It'll keep retrying and will eventually try other pods if there are replicas available. To the end-user for clusters with more than one worker node, it appears as though there are intermittent connectivity problems and performance issues. 

For example - if I have a self-hosted docker registry in my cluster at `registry.example.com` and I have a pod that builds and pushes docker images - if my docker-build pod tries to login to the registry via `registry.example.com` and both pods are on Node A, it won't work and the client will eventually see an nginx error on their web browser. However, if there is a replica on Node B, it will eventually try forwarding the traffic to the pod on Node B. For the user, it feels like the application is slow.

pgagnon:
is `externalTrafficPolicy` set to `Local` or `Cluster`?

cmarteepants:
It would have been whatever the default is. Client resolved it themselves by removing the nlb annotation and preferred using classic. I didn’t push further

jwitz:
[...]

PR #398, opened 2021-06-23
Added a note about when to use ELB instead of NLB

jwitz:
Closes https://github.com/astronomer/docs/issues/386

In addition to adding the ELB note, I corrected some places where we still refer to ELB as the default load balancer. Only updating the v0.25 version of this doc because it's for installation.


paolaperaza (approved):
This LGTM! If @fritz-astronomer or @cmarteepants are cool with this and think this messaging gives enough clarity on this issue, do feel free to merge :)

fritz-astronomer (approved):

cmarteepants (approved):
LGTM 👍 

I agree with @pgagnon suggestion to modify the actual chart, but this at least gives some sort of indicator until we are in a position to implement his suggestion

pgagnon:
We couldn't implement my suggestion in the chart. That would be a documentation fix as well. I think describing both alternatives is best though so OK merging this and expanding later.

pgagnon (approved):

PR #410, opened 2021-07-06
removed terraform from docs

virajmparekh:
Removing the AWS terraform until further notice


paolaperaza (approved):
Merging! Thanks @virajmparekh.

cc: @jwitz FYI the Terraform install docs were the last in the `01_aws` menu so we didn't need to change the numbered file names here (e.g. there was no `03_doc` to rename `02_doc`).

PR #431, opened 2021-08-10
Correct helm chart step reference for aws install guide versions 0.23 and 0.25

### LINKED ISSUE #430: AWS install instructions incorrect reference in Step 2
**Affected Document:**
https://www.astronomer.io/docs/enterprise/v0.25/install/aws/install-aws-standard
  
**Severity:**

- [ ] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [ ] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [x] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**
In Step 2, there is a note that says "Given Astronomer's default resource request of ~11 CPUs and ~40GB of memory, we recommend using either six m5.xlarge or three m5.2xlarge instances for your cluster. To modify Astronomer's default resource requests, see step 6." But Step 6 is about retrieving SMTP URI, not related to default resource requests. 

I believe this should reference **step 8** on configuring the helm chart which is where you would make this change. Although, step 8 doesn't explicitly refer to modifying resource requests, so I think the language could also be taken out entirely if we think that would be more clear (i.e. remove this sentence: "To modify Astronomer's default resource requests, see step 6") 

**Additional Notes:**

collinmcnulty:
fixes #430 

In docs for version 0.23 and 0.25 there are additional steps, so the referenced "step 6" is now step 8.


jwitz (approved):
Great catch @kentdanas, and thank you for getting this sorted @collinmcnulty ! Merging now

Issue #358, opened 2021-05-18
external secrets backend on astronomer cloud doc is misleading


krisdock:
**Affected Document:**

https://www.astronomer.io/docs/cloud/stable/customize-airflow/secrets-backend
  
**Severity:**

- [ ] Critical: This document has incorrect or missing information. Add the `error/critical` label to this issue.
- [x] FYI: The document is misleading or could benefit from additional context or guidance. Add the `error/FYI` label to this issue.
- [ ] Typo: Add the `error/typo` label to this issue.
  
**Suggested Change (include links to relevant existing docs, support tickets, messages, or external content):**

we call out gcp secret manager as a backend secrets tool that's configurable, but then we don't provide instructions on how to set it up despite doing so for the other tools on the list. we should improve the document and provide the instructions for configuring gcp secret manager.

**Additional Notes:**

https://astronomer.zendesk.com/agent/tickets/4660

jwitz:
@krisdock Starting on this using this Notion doc as reference https://www.notion.so/astronomerio/How-to-enable-Google-CloudSecretManager-as-secret-backend-2104169da0e048eaa41006b0859af883#34d25f8072504205b2bcefb76f6e66e3, let me know if there's more considerations here

PR #438, opened 2021-08-24
Added steps for configuring Google Cloud Secret Manager

jwitz:
Resolves https://github.com/astronomer/docs/issues/358 

Used [this notion doc](https://www.notion.so/astronomerio/How-to-enable-Google-CloudSecretManager-as-secret-backend-2104169da0e048eaa41006b0859af883#34d25f8072504205b2bcefb76f6e66e3) as a starting point, so adding @wolfier for technical review. Made a few assumptions, such as the specific role to add and the key file format, that should definitely be reviewed.


paolaperaza (approved):
Thanks for this this, @jwitz ! LGTM. The only thing I'd double check on here is whether or not we need the `ENV` in Env Vars when you add them via the Astro UI.

Otherwise feel free to merge whenever you're ready ✅

PR #469, opened 2021-09-23
Fix query on Type Deployment which contains an id, not uuid

BasPH:
One of the queries in the [Houston API docs](https://www.astronomer.io/docs/cloud/stable/manage-astronomer/houston-api#sample-queries) failed because the type Deployment holds a field `id`, not `uuid`.

Tested updated query on a 0.25.4 installation, and checked source code for older versions. There used to be a field `uuid` for Deployment in older versions, but since `id` was also there, I think we can safely update the query for all versions (v0.12.0 code [here](https://github.com/astronomer/houston-api/blob/v0.12.0/src/schema.graphql#L201-L202)).


enterprise/v0.12/06_manage-astronomer/04_houston-api.md
```diff
@@ -82,7 +83,7 @@ The `workspaceDeployments` query requires the following inputs:
 and can return any of the fields under `Type Details`:
 
 - `config`
-- `uuid`
+- `id`
```
paolaperaza, line 86: @BasPH I see that `workspaceUuid` is still used as part of the pre-requisites. Should we remove all instances of `uuid`, at least where you already replaced them in the example query/mutation?

Seeing the same for most of the examples below.

enterprise/v0.12/06_manage-astronomer/04_houston-api.md
```diff
@@ -159,7 +160,7 @@ mutation DeleteDeployment {
   deleteDeployment (
     deploymentUuid: "<DEPLOYMENTUUID>"
   ) {
```
paolaperaza, line 162:
```suggestion
    deploymentId: "<DEPLOYMENT_ID>"
  ) {
```

For example, should this be changed?

BasPH: The `deleteDeployment` mutation takes a required argument `deploymentUuid` ([source](https://github.com/astronomer/houston-api/blob/main/src/schema/mutation.js#L214)), so that remains the same.

enterprise/v0.12/06_manage-astronomer/04_houston-api.md
BasPH: I checked all other places where `uuid` is used in this doc, but it appears this is the only case where it should be replaced by `id`.

I checked the source and from what I understand all queries take an argument `somethingUuid`, which is then mapped to a field `id`. So the `workspaceUuid` in this query is still correct, but the request for the underlying field `uuid` is not because it doesn't exist, and it should be replaced by `id`.

PR #491, opened 2021-10-13
Enterprise v0.26 release notes + doc changes

jwitz:
Release notes and relevant doc changes for both Enterprise v0.26 and AC 2.2.0. There's a lot of changed files here, but only a few are actually new - the rest are because we cut a new minor docset version. That means you only have to review:

- `/enterprise/v0.26/10_resources/01_release-notes.md`
- `/enterprise/v0.26/04_deploy/02_configure-deployment.md` (For Triggerer information)

There are still some things missing from this PR, mainly:

- Upgrade instructions from v0.25 > v0.26
- LTS support information/ updates

PR #495, opened 2021-10-19
0.26 doc folder maintenance

jwitz:
This PR handles the messy work for 0.26, including updating hyperlinks, directories, and version compatibility tables. It doesn't include content changes related to v0.26 changes, which are available for review in https://github.com/astronomer/docs/pull/491/files


cloud/stable/05_manage-astronomer/02_houston-api.md
```diff
@@ -30,15 +31,15 @@ In short, the Playground is a portal that allows you to write GraphQL queries di
 
 ### Navigate to the GraphQL Playground
 
-On Astronomer Cloud, the URL at which you can reach Houston's GraphQL playground is: https://houston.gcp0001.us-east4.astronomer.io/v1
+On Astronomer Cloud, the URL at which you can reach Houston's GraphQL playground is `https://houston.gcp0001.us-east4.astronomer.io/v1`.
```
danielhoherd, line 34: Shouldn't this be a hyperlink instead of a code block? Not sure if the short-form markdown hyperlink syntax works in our docs env but it would be:
```suggestion
On Astronomer Cloud, the URL at which you can reach Houston's GraphQL playground is <https://houston.gcp0001.us-east4.astronomer.io/v1>.
```

jwitz: I did some formatting so that this link wasn't naked, but you can still click on it and get to the playground.

PR #517, opened 2021-11-16
Use md headings instead of bold text. prettier.

danielhoherd:
It seems like markdown headings are a better mechanic to use in the template, if only because it's a cleaner look, but also because that's what headings are for. If this isn't what you all want feel free to close it.

I also ran `prettier -w`. You all might want to put that in .pre-commit or something because there are other syntactic weirdnesses in this repo, like trailing whitespace.


jwitz (approved):
LGTM @danielhoherd . I went with bold originally because we had some issue writers mistakenly add their text to the end of the header, which then looks like one big title when rendered. But over time, it seems like most issue writers have a firm grasp of Markdown, so I'm cool with this!

danielhoherd:
I hadn't thought of that problem, but I definitely see how folks could do that. One thing I love about using proper headings in markdown files on GH though is that you get a TOC, but that's probably not super important for these issues. It's really great in READMEs.

PR #525, opened 2021-12-02
Demonstrate Helm install using an existing version

BasPH:
Astronomer Enterprise `0.26.2` doesn't exist, I suggest to demonstrate commands using an existing version `0.26.4`.


jwitz (approved):
Nice catch! Wasn't too sure what patch version we'd end up making public when I wrote this. Thanks @BasPH

